PR #5, opened 2018-12-13
Modifications for a simple phased array.

daniel-zid:


cg-laser:
thanks for starting to work on a better phased array trigger. However, it does not yet do what you want. You just add up all the traces, but the time delays of all the channels are different. 
A first approximation would be to shift all traces by the signal time we know from MC (although this is already a little tricky because we have two times for the two ray tracing solutions, so we first need to figure out which one caused the stronger signal). This overestimates the sensitivity because it assumes that we always have a perfect beam, but a good first step.

daniel-zid:
You are right. That's what I called it simple trigger and set all the
cables delays to the same value. It's a veeery preliminary solutions for
giving some results to Jakob in a few days.

Do you want to write a trigger class that controls the actual phasing,
regardless of the cable delays?

Le jeu. 13 déc. 2018 à 19:24, Christian Glaser <notifications@github.com> a
écrit :

cg-laser:
I think these results will be relatively useless because if the phase delays are not correct, you might not get any signal amplification. Why don't you do the following what we also did for the proposal: Estimate the phased array with a single antenna at a 2 sigma single sided threshold? Then you can just use the simpleThreshold trigger module. 

Regarding your second question: We don't need a fancy trigger class, but the trigger module needs to do this.

daniel-zid:
But what kind of phasing do we want? Relative delays of zero are equivalent
in principle to pointing the array towards the theta = 90 deg direction. If
we use a single antenna and reduce the threshold that's not realistic
either because for some arrival directions you will actually see less with
a phased array.

Le jeu. 13 déc. 2018 à 20:15, Christian Glaser <notifications@github.com> a
écrit :

cg-laser:
Therefore, the phased array form multiple beams to get a full zenith 
coverage. But as you say, not all arrival directions will have the same 
sensitivity. However, this is all studied in detail in the phased array 
paper, and they find that they get on average 2 sigma.

The 2 sigma was with 7 antennas, with 8 antennas you can probably get 
1.8 sigma, and with further hardware improvements they even claim to get 
down to 1.5 sigma, but the event quality is questionable.

Am 13.12.18 um 11:34 schrieb daniel-zid:

daniel-zid:
Ok, I'll use a single antenna if you think that's the best solution. I will
also take a look at the paper. Can you give me the reference to make sure
that it's the same that I already have, please? Thanks

Le jeu. 13 déc. 2018 à 21:06, Christian Glaser <notifications@github.com> a
écrit :

anelles:
Have we paused this or is there some status change?

daniel-zid:
Paused. I'm following Christian's advice and perform all the calculations
for the white paper
with a single antenna and a lowered threshold. Once everyone is happy with
the simulations
for the Gen2 paper we can study the effect of a realistic phasing. For the
aperture calculations,
it should not matter that much.

PR #6, opened 2018-12-17
Prevent uneven trace samples

christophwelling:
-Prevents time traces from having an uneven number of samples.
-Makes the get_trace and get_frequency_spectrum functions set the _time_trace or the _frequency_spectrum to None if it is out of date, since they are not used and only make files larger.

### DIFF
--- a/NuRadioReco/framework/base_trace.py
+++ b/NuRadioReco/framework/base_trace.py
@@ -53,6 +53,12 @@ class BaseTrace:
         sampling_rate: float
             the sampling rage of the trace, i.e., the inverse of the bin width
         """
+        if trace is not None:
+            if trace.shape[trace.ndim - 1]%2 != 0:
+                if trace.ndim == 1:
+                    trace = np.append(trace, 0)
+                else:
+                    trace = np.append(trace, np.zeros((trace.shape[0], 1)), axis=1)
         self.__time_domain_up_to_date = True
         self._time_trace = trace
         self._sampling_rate = sampling_rate
--- a/NuRadioReco/framework/base_trace.py
+++ b/NuRadioReco/framework/base_trace.py
@@ -31,6 +31,7 @@ class BaseTrace:
 #             logger.debug("time domain is not up to date, calculating FFT on the fly")
             self._time_trace = fft.freq2time(self._frequency_spectrum)
             self.__time_domain_up_to_date = True
+            self._frequency_spectrum = None
         return self._time_trace
 
     def get_frequency_spectrum(self):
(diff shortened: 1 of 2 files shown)


cg-laser:
it might be dangerous to append a zero to the trace because this introduces an artificial high frequency component. I think it would be better to cut away one sample. 
The alternative would be to just fail if the trace is set with an odd number of samples. Right now I think this is the solution, as odd traces should never happen. 
Can you remind me how you ran into this problem? I thought the code is written such that only traces with even number of samples are created? It might be better to fix the modules that result in odd traces.

christophwelling:
I did not think about the risk of adding an artifact to the frequency spectrum, but I agree that it is a problem.
So far I have only seen uneven numbers of samples in the E-fields, not in voltage traces. The way I am reconstructing events these are only affected by the voltageToAnalyticEfieldConverter and the stationResampler, but I have not found anything suspicious in these modules. There is of course also the chance that something goes wrong when files are saved and loaded.

anelles:
I agree that adding a sample creates a problem. There is of course a trace length at which is doesn't matter, but if you consider that the cutting module could cut the pulse to just 10 samples or so (around the pulse) you wouldn't want to add 10% of data. 
I find it indeed curious that things should happen in the efield converter. 
Can we maybe add a warning to prevent the trace-length ever being uneven?

cg-laser:
yes this would also be my method of choice, just raise an exception if 
an uneven trace length is set and fix the modules. I'm also puzzled how 
the efield converter can create odd trace length.

Am 17.12.18 um 10:22 schrieb Anna Nelles:

anelles:
Ok, Christoph, do you want to change this as discussed or do you disagree?

christophwelling:
Yes, this seems like a good solution, but I think we should first figure out where these uneven trace lengths come from. Otherwise it will just make the efield-converter (or whatever module is causing this) unusable.

cg-laser:
First: The main reason why I suggested to raise an exception when a odd tracelength is set is to find out where this happened (the stack trace will tell you). 
Second: the 'get_array_of_channels' function does to following important task: It does not only shift the trace times according to a plane wavefront, but it also takes into account different cable delays (or in general any time shift between channels stored in the trace start time). This is needed for all efield reconstruction modules. 
Because our traces are often very short, rolling the traces is not an option as a signal from the beginning might end up at the end of the trace. In general, I'd say that rolling is always physically wrong and only applicable in simulation studies where we only have a single pulse and the rest of the trace is zero. 
So what is done instead is that all traces are cut to the overlapping region of same physical time which will always reduce the tracelength. 
For a cosmic ray stations with 4 LPDAs with exactly the same cable length, just rolling the trace might be an option, keep in mind that these modules should also work for more complex stations with differences in cable delays (>100ns) that are similar to the trace length. 

Before the efield reconstruction, the 'channelStopFilter' should be applied to filter out the stop glitch and to extend the traces with zeros. (see https://github.com/nu-radio/NuRadioReco/wiki/Times for more details). 

As you also say @christophwelling, the function has a saveguard to not produce odd tracelengths. So how can it be the cause for the bug?
Your say it is being downsamled later? Where? Downsampling should also preserve the 'even' property.

cg-laser:
oh I see, good point, yes downsampling can of course lead to odd trace lengths... then I suggest to catch it in the stationResampler.py module as in the channelResampler.py module. 
And also to raise an exception if someone tries to set a trace with an odd number of samples.

PR #7, opened 2019-01-04
Eventbrowser voltage from rec trace

christophwelling:
Changes 2 things:
1. adds 2 utility functions. One returns the antenna response for a give incoming direction. This was already part of the voltageToEfieldConverter's  get_array_of_channels function but was moved into its own utility function for easier use. The other returns the voltage traces that would result from the reconstructed electric field.
2. adds the option to also show the voltage traces that would result from the electric field in the voltage trace plots of the event browser.

### DIFF
--- a/NuRadioReco/eventbrowser/apps/traces.py
+++ b/NuRadioReco/eventbrowser/apps/traces.py
@@ -466,7 +466,7 @@ def update_time_traces(evt_counter, filename, dropdown_traces, dropdown_info, st
         channel_ids = []
         for channel in station.iter_channels():
             channel_ids.append(channel.get_id())
-        for i_trace, trace in enumerate(trace_utilities.get_channel_voltage_from_efield(station, channel_ids, det, station.get_parameter(stnp.zenith), station.get_parameter(stnp.azimuth), antenna_pattern_provider)):
+        for i_trace, trace in enumerate(trace_utilities.get_channel_voltage_from_efield(station, channel_ids, det, station.get_parameter(stnp.zenith), station.get_parameter(stnp.azimuth), antenna_pattern_provider, cosmic_ray_mode=station.is_cosmic_ray())):
                 fig.append_trace(go.Scatter(
                     x=station.get_times()/units.ns,
                     y=fft.freq2time(trace)/units.mV,
--- a/NuRadioReco/eventbrowser/apps/traces.py
+++ b/NuRadioReco/eventbrowser/apps/traces.py
@@ -11,14 +11,21 @@ from app import app
 import dataprovider
 from NuRadioReco.utilities import units
 from NuRadioReco.utilities import templates
+from NuRadioReco.utilities import trace_utilities
+from NuRadioReco.utilities import fft
+from NuRadioReco.detector import detector
 from NuRadioReco.framework.parameters import stationParameters as stnp
 from NuRadioReco.framework.parameters import channelParameters as chp
+import NuRadioReco.detector.antennapattern
 import numpy as np
 import logging
+
 logger = logging.getLogger('traces')
 
 provider = dataprovider.DataProvider()
 template_provider = templates.Templates()
+det = detector.Detector()
+antenna_pattern_provider = NuRadioReco.detector.antennapattern.AntennaPatternProvider()
 
 layout = html.Div([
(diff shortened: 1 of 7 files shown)


anelles:
Looks good to me. As long as this doesn't get too heavy on the on-the-fly calculation side of things on the eventbrowser. But this I cannot judge.

cg-laser:
good idea to move the functions to a general place. 

However, looking through the code again I found one problem (that was already there before). The function automatically determines if the zenith angle needs to be refracted into/out of the ice. The refracting into the ice makes only sense for cosmic rays but not for neutrinos. We often have downward coming signals that still originate from within the ice. Therefore, I added a 'cosmic_ray' mode in the 'efieldToVoltageConverter.py'. The default is that is works for neutrinos. 
I think we should add another argument to the `get_efield_antenna_factor` function `cosmic_ray_mode=False`. Only if cosmic_ray_mode=True, refraction into the ice is taken into account. (as implemented in  'efieldToVoltageConverter.py')

christophwelling:
@anelles The new plot is not drawn by default and has to be selected from a dropdown to be seen, so it is not going to impact performance unless you actually want to see it.
@cg-laser I added the cosmic_ray_mode parameter where it is needed. I agree that this change makes sense, but the screenshot shows a new problem this creates: Since the event browser has no way of knowing that this is a cosmic ray event, it does not take the refraction into account.
I think a cosmic ray/neutrino flag at the station level (maybe in the parameter storage?) could fix this and would be pretty useful in general. During event reconstruction we have to make assumptions about what kind of event we are dealing with anyway, and storing this somewhere would make a lot of stuff simpler. For example it would remove the need to set the cosmic_ray_mode parameter manually in each module.

eb_voltage_from_efield

cg-laser:
@christophwelling good point. Yes a cr/nu flag does make sense. How do we implement it? 
I think we should define a member variable in station.py and implement the functions `is_cosmic_ray` and `is_neutrino`. This allows to correctly initialize the variable and I would recommend to set it to 'neutrino' by default. 
Then we would need a new module that activates the cosmic ray mode `cosmicRayModeActivator`. 
What do you think?

christophwelling:
I agree, with 2 caveats:
1. The flag should be in base_station.py because it is also useful to have in a sim_station. The use case I see here is that we may for example have a simulated cosmic ray event, but our analysis misidentifies it as a neutrino. In principle you could of course just try and remember what you simulated, but a flag would make things much easier.
2. I don't think a new module is necessary. For the sim_station the logical place to set the flag would be the CoREAS reader or in NuRadioMC itself. For the reconstruction this should be done by the channelTemplateCorrelation. Depending on what sort of template gets a good match, the flag is set accordingly.
In the long term we probably need a dedicated module to identify cosmic rays (for background rejection) that would then take over this job, but for now I think the templateCorrelation is the logical place to do this and a new module is not necessary.

cg-laser:
yes we should have it in base_station. 

Regarding the second point: There are many use cases where you want to force the templateCorrelation module to use either cosmic ray or neutrino templates. So we can't just let this module decide. 
I think the user needs to be capable to force a certain event type. But this module does not need to be necessarily used in all pipelines.

christophwelling:
I added the is_cosmic_ray and is_neutrino functions as well as the set_is_neutrino function to set the flag. If a new station is created, the flag is set to neutrino by default. When a CoREAS event is read, the CoREAS readers set the flag to cosmic ray for the sim station (but not the station).
To set the flag during reconstruction I added a new module called cosmicRayIdentifier (in principle it could also be used on sim_stations, but there is probably no good reason to do so). The idea is that this module is responsible for deciding if an event (or more specific a station) should be treated as a cosmic ray or neutrino. Which criteria this decision should be based on can be specified using the 'mode' parameter of the run function. For now, the only mode it supports is 'forced' which just sets all stations to cosmic ray, but once we start implementing methods distinguish between cosmic rays and neutrinos (e.g. signal in upward vs. downward pointing antennas) these can be added as another mode.

christophwelling:
The cosmic ray/neutrino flag is currently not used in the efieldToVoltageConverter or the voltageToEfieldConverter. I would prefer to make these changes along with the new electric field class, since some of these modules changed quite a lot and merging later would probably not be a lot of fun.

PR #10, opened 2019-01-11
Efield object

christophwelling:
Adds electricField class to store electric fields. They can be accessed by station.get_electric_fields and station.get_electric_fields_for_channels. Each electricField object stores a list of IDs of the channels it was reconstructed from, or for which it was simulated as well as it ray_path_type (for simulated E-fields). 
The individual modules can mostly be used as before, with the following changes:
- Station traces and sim_station channels should not be used anymore to store electric fields.
- stationResampler, stationBandpassFilter and stationSignalReconstructor have been renamed to electricFieldResampler, electricFieldBandpassFilter and electricFieldSignalReconstructor.
- efieldToVoltageConverter has been removed. Instead, the efieldToVoltageConverterPerChannel has been renamed to efieldToVoltageConverter and can now also handle cosmic rays.
- The cosmic_ray_mode parameter has been removed from several modules. The station's is_cosmic_ray() flag is used instead.


cg-laser:
For documentation: A discussion that we had via email before starting this pull request:

Hi,

I have added a zenith and azimuth parameter to the E-field parameter storage for now, so E_theta and E_phi are well-defined. As I wrote previously, if we want to we can later overwrite the set_trace and get_trace functions. I agree that a coordinate system tool would be very useful, but I would like to finish the changes to the electric field first. Messing around with that is already disruptive enough, so I would not want to make it more complicated by changing the coordinate system at the same time. We can then work on the coordinate system once we are reasonably sure that the new E-field works fine.

About not saving more than we need, I want to point out that right now, the electric field has an E_r component. It is of course 0, but it exists and takes up storage. Should we get rid of it? Unfortunately, I don't think the cs transformations in the radiotools can handle that right now, so maybe wait until coordinate system tools are implemented in NuRadioReco?

As for what these tools should look like, how about some vector class like it is done in Offline? That seemed to work pretty well (at least for me).

I also came across another detail that I am not quite happy with. If we assiciate E-fields with channels, we imply that these are what the field (and the signal direction) is at the antenna position, but this is not true for cosmic rays because of refraction at the air/ice boundary. It is not a big problem, but I can see how this could easily confuse people, so if anyone has an idea about how to clear this up, please tell me.

Cheers,

Christoph

cg-laser:
don't we want to remove the channels from sim_station? Or is it needed for anything?

anelles:
I like the suggestion to use this opportunity to rename the default to .nur files. This will make it transparent, that we are breaking backwards compatibility with this change.

christophwelling:
Update:
- parameter indices reset to what they were originally
- CoREASreader gets detector in the run() method. Channel-IDs are taken explicitly from the detector description.
- SimStation channels removed. They are indeed not needed for anything anymore (I just forgot to remove them because I assumed they were inherited from BaseStation)
- While I was on it I also removed Station.get_channels(). That one had a deprecation warning, so now is a good time to follow through with that.
- About file endings: I think currently the only place in NuRadioReco where this is relevant is in the Eventbrowser, which will only look for .ari files (I now changed that to .nur). The eventWriter doesn't enforce any file endings. We may want to change this and just add a .nur to the filename if it is missing. This way we could also get rid of the eventWriter's annoying habit of producing files with names ending in things like .ari_part02
- We talked about not wanting to store the E_r component of electric fields. I would suggest to overwrite the get and set functions for the traces and spectra. set_trace would check if the passed array has 2 or three components and only save E_theta and E_phi. The getter function would add the E_r component again, unless told not to do so (something like get_trace(n_dim=3), so it can be used the same way as before)
- Increasing the version number sounds good. Is that done using the VERSION variable in modules/io/ARIANNAio.py?

anelles:
I am all in on making the EventWriter obey to .nur. And at least have the EventReader throw a warning, if an non .nur file is read in. Any thoughts, @cg-laser Christian?

cg-laser:
I think not saving E_r is not so critical but if it can be done transparently that the missing dimension is added on the fly in any getter function, I'm ok with it. We could even just do it in the (de)serialization function to save disk space. As we anyway don't consume a lot of memory, we could always have the three dimensions in memory. 

how would you implement splitting files up into chunks without adding something like "part_02" at the end of the filename? The ability to split up the datafile into chunks of e.g. 1GB is a very useful feature. 

And yes you need to change it in "modules/io/ARIANNAio.py".

cg-laser:
and fine for me to enforce the *.nur file ending in the eventWriter. Then the user only needs to specify the first part of the filename but not the extension. @christophwelling how do you suggest to split up events into multiple files.

christophwelling:
I think adding _part02 to the filename is fine, it just looks weird that it is part of the file extension. I think "filename_part02.ari" makes more sense than "filename.ari_part02". Other than that I think it's working fine.

christophwelling:
Update:
- File/software version set to 2.0
- eventWriter adds .nur to file name ending (unless the file already ends with .nur). If the user tries to save a file as .ari, it is changed to .ari.nur and a warning message asks the user to please use .nur next time

cg-laser:
I leave this pull request open for a while in case we still need to add something.

cg-laser:
this brings up actually another general problem: the signal time of the efield. For neutrinos, each efield has a start time that corresponds to the travel time from the vertex to the particular channel. 
For cosmic rays we just have one efield for all channels and assume time zero is at the station center...
Given our new efield structure the reference point of the time is not properly defined. If we have only one corresponding channel, it is this channel, but what if we have several? We could add a 'reference point' property to the efield class which stores the position (xyz) relative to the station center. But for the cases with only one channel this will be cumbersome but doable.

Then the efieldToVoltageConverter can calculate additional delays from the displacement of the channel to the efield reference point if necessary. 
Line 96 is currently `t0 = electric_field.get_trace_start_time() + cab_delay` and would need to be replaced by 
`t0 = electric_field.get_trace_start_time() + cab_delay + displacement_delay` 
and then again in 147

PR #11, opened 2019-01-16
Efield object

christophwelling:
- When converting simulated E-fields to voltages, cable delay is now treated in the efieldToVoltageConverter only.
- The efieldToVoltageConverter takes different signal travel times for cosmic ray signals into account


anelles:
I am confused. Hasn't most of this already been merged by Christian in the other pull request?

christophwelling:
Yes, but the efieldToVoltageConverter didn't work for cosmic rays. This pull request is for that fix

anelles:
It says here that you want to merge 37 commits ...

christophwelling:
Apparently it also counts the commits on the branch that were already merged. It really is just one commit

christophwelling:
I took this over from how it was implemented in the "original" efieldToVoltageConverter (the module now called efieldToVoltageConverter used to be called efieldToVoltageConverterPerChannel). I assumed that whoever implemented it like that had a good reason to do it. But yes, it seems weird.

cg-laser:
Didn't see that you opened a new pull request. I thought one can just reopen the previous one. 

this brings up actually another general problem: the signal time of the efield. For neutrinos, each efield has a start time that corresponds to the travel time from the vertex to the particular channel.
For cosmic rays we just have one efield for all channels and assume time zero is at the station center...
Given our new efield structure the reference point of the time is not properly defined. If we have only one corresponding channel, it is this channel, but what if we have several? We could add a 'reference point' property to the efield class which stores the position (xyz) relative to the station center. But for the cases with only one channel this will be cumbersome but doable.

Then the efieldToVoltageConverter can calculate additional delays from the displacement of the channel to the efield reference point if necessary.
Line 96 is currently t0 = electric_field.get_trace_start_time() + cab_delay and would need to be replaced by
t0 = electric_field.get_trace_start_time() + cab_delay + displacement_delay
and then again in 147

anelles:
Ok, let's add a warning and maybe put a pin in it (i.e. open an issue that we don't forget) and then we can merge this in.

christophwelling:
Yes, we need a proper definition of what the E-field times are relative to. I think this will also be relevant for neutrinos. At some point we will reconstruct E-fields using different antennas. We will probably end up reconstructing separate E-fields for the surface and the deep detector components or direct vs. reflected signal paths, and then we also need a proper definition of their timings.
But I think we should do that in a separate branch. Things are working fine for now and we are not blocking us of from fixing this later.

PR #13, opened 2019-01-21
Coreas selection

anelles:
I have added a module that removes CoREAS sims at the first step after the read in. This avoids all simulations that don't have signal in the interesting band, i.e. < 50 MHz only.
Not a fancy module, but it does do the trick most of the time.

### DIFF
--- a/NuRadioReco/modules/channelTemplateCorrelation.py
+++ b/NuRadioReco/modules/channelTemplateCorrelation.py
@@ -12,6 +12,7 @@ from NuRadioReco.framework.parameters import channelParameters as chp
 import logging
 logger = logging.getLogger('channelTemplateCorrelation')
 
+import matplotlib.pyplot as plt
 
 class channelTemplateCorrelation:
     """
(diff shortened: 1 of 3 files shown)


anelles:
Hi Christian, thanks for checking:
@1: I have changed to logic for upsampled traces. The issue with significant signal is not really a problem, since the traces will then always be called good. I have checked this case.
@2: Well you want the trace as soon as the it has any frequency above the minumin frequency. The window doesn't really matter, no? So the logic like this give the trace the best chance.
@3: Has been added.

anelles:
Hi, @cg-laser, I think this now does everything that you asked. Can I merge? 
I also just accidentally committed a fix for the templateCorrelator to this branch. This is just sorting the logic a bit more reasonable if you use n> 1 templates.

cg-laser:
@2: it can happen that the maximum frequency bin is outside of the interval you are considering. Then the trace would be marked as unusable event it is has sufficiently high amplitudes within the band.

cg-laser:
I committed a change to the selection that should solve my concern with point 2. @anelles can you check if this makes sense? If yes we can merge.

anelles:
I actually think they both do the trick, but yours is more elegant.

PR #14, opened 2019-01-22
Makes voltageToAnalyticEfieldConverter use the slope from the res_obj…

christophwelling:
It makes more sense to use the slope that was fitted in the obj_xcorr fit as a default for the other fits instead of some hardcoded value.
I tested it on a few events, I could not see any change on the results. Still, I think if we have a rough estimate of what the slope is, we should use it instead of just guessing.


anelles:
Fine. A random guess is probably not a good idea. 
Do we want to add a sanity check, whether the slope is something reasonable, i.e. that you don't give it some sort of diverging starting value? Not sure what the obj_xcorr values can be.

cg-laser:
thanks for spotting this, I thought we already used the slope from the first fit in the amplitude fit. 
I agree with Anna that a rough sanity check would make sense though.

PR #15, opened 2019-01-28
save ARIANNA specific parameter 

cg-laser:
For ARIANNA analyses we need access to ARIANNA specific parameters such as the sequence number and sequence start and stop time. I think the best way to save these additional parameters is in a dedicated parameter storage class. It only adds up to the data volume if it is filled. 
The only downside is that we have ARIANNA specific code in one of the base classes. What do you think?


anelles:
I already stumbled over the same thing. Having ARIANNA specific parameters as their own class is fine I guess.
Christian, there are lots of other parameters in the files that have never made it to snowshovel, such as trigger settings. Do we also need to do something about this? 
(Maybe this is not the right thread, but for neutrino searches, you do need the right thresholds for example)

cg-laser:
I already added the thresholds ;-). I would add parameters on demand... 
It is just too much work to replicate all the snowshovel information to 
the NuRadioReco data format now and most likely we won't be using many 
of the parameters. So whenever someone needs something for analysis, I 
would add it to the reader.

Am 29.01.19 um 00:07 schrieb Anna Nelles:

PR #16, opened 2019-01-29
Noise importer fix

christophwelling:
Fixes errors in the noiseImporter that were caused by the changes in the event structure

### DIFF
--- a/NuRadioReco/modules/channelLengthAdjuster.py
+++ b/NuRadioReco/modules/channelLengthAdjuster.py
@@ -1,6 +1,7 @@
 import numpy as np
 import copy
 import logging
+import datetime
 from NuRadioReco.utilities import units
 logger = logging.getLogger('channelLengthAdjuster')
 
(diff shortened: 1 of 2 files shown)


cg-laser:
@christophwelling can you remind me why the 'relative station time' does not exist anymore? For absolute timing we need to keep track of changes in the trace length.

christophwelling:
I don't think I saw the relative station time ever used before, so I have no idea. But wouldn't the logical place to keep track of this be the channel anyway?

cg-laser:
according to https://github.com/nu-radio/NuRadioReco/wiki/Times we have the station start time to keep track of time changes that effect all channels in the same way, and the channel_start_time to do the same time shifts different for each channel

cg-laser:
I just went through the history but couldn't find the commit that removed the relative station time, hence we must have removed it before we transferred it to github...  I'm still puzzled why this quantity is gone... As station inherits from base_trace, the set_trace_start_time would be available.
BTW: Do we need the inheritance of base_station from base_trace anymore now that we have the efield class? I don't think so. 

We definitively need to keep track of the time changes, but it is probably better to do it for each channel separately.

PR #18, opened 2019-01-31
io modifications

cg-laser:
apply 'micro' mode for efields: Don't save efields if write mode is 'micro'. 

@christophwelling Can we remove the inheritance of base_trace from the base_station class?


christophwelling:
Yes, I think inheriting from base_trace is no longer necessary. None of the functions in the base_trace class make sense for a station with the new event structure

cg-laser:
I removed the base_trace inheritance and renamed the ARIANNAio class to NuRadioRecoio

PR #20, opened 2019-02-08
Example

iplaisier:
example reconstruction cycle


cg-laser:
@iplaisier thanks for adding this example!

I thing we should add commentary to the code such that the example is understandable on its own. 

The indentation is not correct starting at line 105

iplaisier:
Together with the wiki page i think it should be doable https://github.com/nu-radio/NuRadioReco/wiki/Example:-Full-reconstruction-cycle. 
Let me know if you think otherwise.

PR #32, opened 2019-03-20
Typo fixes

anelles:
Somewhere in previous commits, there were a couple of typos introduced. 
Also, we should use the method that works for both lists and arrays.

### DIFF
--- a/NuRadioReco/modules/io/coreas/simulationSelector.py
+++ b/NuRadioReco/modules/io/coreas/simulationSelector.py
@@ -81,8 +81,9 @@ class simulationSelector:
             noise_std = np.std(noise_region)
 
             noise += n_std * noise_std
-            
-            mask = (freq >= frequency_window.min())  & (freq <= frequency_window.max())
+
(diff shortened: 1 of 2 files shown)


NuRadioReco/modules/io/coreas/simulationSelector.py
```diff
@@ -81,8 +81,9 @@ def run(self, evt, sim_station, det, frequency_window = [100*units.MHz,500*units
             noise_std = np.std(noise_region)
 
             noise += n_std * noise_std
-            
-            mask = (freq >= frequency_window.min())  & (freq <= frequency_window.max())
+
+            mask = np.array((freq >= np.min(frequency_window))  & (freq <= np.max(frequency_window)))
```
christophwelling: I don't think the np.array command is really necessary here

anelles: It is if you give it a list for the frequency window as the documentation suggests.

PR #44, opened 2019-04-12
Some small changes to the traces plot in the eventbrowser to make it …

christophwelling:
Small changes to the traces plot in the eventbrowser to make it usable with a large number of channels


cg-laser:
why did you remove the plotting of RMS and L1 values. These are very useful quantities if you look at data. The plotting of these quantities was anyway optional, so I don't see a reason why to remove them.  (and it would be good if the commit message, or the pull request, was more explicit about the changes).

anelles:
@christophwelling this is now in conflict given more recent changes. Any updates?

christophwelling:
Okay, I didn't realize that these annotations were still used by people. They are back now.
Now all it changes is that the height of the individual channel traces/spectra plot is set depending on the number of traces, so that they don't get squished when we have a lot of channels.
Merge conflicts have been resolved.

cg-laser:
thanks, looks good to me!

cg-laser:
Thanks, looks good to me

PR #45, opened 2019-04-18
antennapattern: 180 degree roation, VtoEconveter: added trace start times.

ggaswint:
hot fix for 180 degree LPDA rotation in antenna models from WIPL'd. Added trace start time to Efields to get right timing delays between channels.


anelles:
@ggaswint are you still working on this? I am pinging all older pull requests.

ggaswint:
Hi Anna,

Sorry I haven’t gotten back to you sooner. I’ve been really busy with my
teaching schedule this quarter. Let me bring this up with Christian today
and get back to you. We mutually decided to push this “hot fix” to
NuRadioReco.

Sincerely,
Geoff

cg-laser:
from the SPICE core measurement we know that the polarity of one antenna type is wrong. It could be either the LPDA or the dipole. As we don't have WIPL-D licence at the moment we can't check. Rotating all antennas (in azimuth) by 180deg will will change the polarity of the LPDAs but not the dipoles. 
But that's maybe not the best thing to do... As this will screw up antenna orientations that are not symmetric under a 180deg phi rotation. E.g. the Vpol LPDA ARIA setup... 
Any other thought how to fix this for now in the code? Or as parameter?

ggaswint:
Christian and I have decided to remove this branch. Instead, to match the polarities between first 4 and last 4 channels of station 51, a 180 degree zenith flip was done in the online SQL database for the bicone antennas that are channels 4-7 of station 51.

PR #46, opened 2019-04-18
Direction reco

cg-laser:
branch improving the directional reconstruction


NuRadioReco/modules/correlationDirectionFitter.py
```diff
@@ -143,6 +146,7 @@ def ll_regular_station_fft(angles, corr_02_fft, corr_13_fft, sampling_rate, posi
 
         if use_correlation:
             # Correlation
+            print(signs)
```
ggaswint: Perhaps this print statement and the one on line 170 should be commented out.

PR #48, opened 2019-05-02
Event browser mini

christophwelling:
- Makes eventbroser usable for files saved in mini/micro format by catching exceptions if traces are missing
- Fixes bug that prevented event selection by clicking on points in certain plots


anelles:
Thanks for fixing the browser!

PR #50, opened 2019-05-03
No highlevel fix

anelles:
CoREAS HDF5 files may not have high-level quantities computed. This now allows for the reader to still work but does not compute the weights. Issues a warning.

### DIFF
--- a/NuRadioReco/modules/io/coreas/readCoREASStation.py
+++ b/NuRadioReco/modules/io/coreas/readCoREASStation.py
@@ -1,8 +1,11 @@
 import h5py
+import numpy as np
 import NuRadioReco.framework.event
 import NuRadioReco.framework.station
 from NuRadioReco.modules.io.coreas import coreas
 
+import logging
+logger = logging.getLogger('readCoREASStation')
 
 class readCoREASStation:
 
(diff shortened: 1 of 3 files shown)


NuRadioReco/modules/io/coreas/readCoREASStation.py
```diff
@@ -21,14 +24,20 @@ def begin(self, input_files, station_id):
...
         self.__input_files = input_files
         self.__station_id = station_id
-        self.__current_input_file = 0        
+        self.__current_input_file = 0
 
     def run(self, detector):
         for input_file in self.__input_files:
 
             # read in coreas simulation and copy to event
+#             if input_file == "../../data/CR_sims/005548.hdf5":
```
christophwelling: Why do we need this line? We should not hard-code this anyway.

anelles: Was commented out already anyway. Removed it for cosmetic reasons now in any case.

cg-laser:
this also means that the coreas reader that randomly assigns core positions already worked with non-highlevel files?

cg-laser:
line 103 in coreas.py also needs attentions

anelles:
No, it just doesn't calculate weights, but no longer crashes. The issue to do this on the fly is still open. I have changed coreas.py in the code cleanup branch.

PR #51, opened 2019-05-03
correlationDirectionFitter update

christophwelling:
There was still code there that assumed the sim E-fields were stored in simStation channels. That code was removed. It was only used for debug output but could cause crashes. The debug output now only runs for CR events, but since the correlationDirectionFitter does not work properly for nu events that should be fine.


cg-laser:
there is a direction-reco branch where this is already addressed and fixed.... maybe leave a little more time to review?

PR #53, opened 2019-05-03
Cleanup

anelles:
Just what crossed my path today in documentation cleaning and centralization.


anelles:
If you tell me the function in radiotools that you would use instead? If it is a 1:1 replacement, I agree. But I cannot find it easily.

anelles:
Please all continue to work on the documentation and commit to this branch. Especially things like detector.py and the eventbrowser need mode documentation.

cg-laser:
maybe not the cleanest thing to do, but I added new functions to retrieve the absolute position from the sql data base identified by a string (e.g. position 'X'). To make it available to the json version, I also needed to export the full positions table to the json file.

anelles:
Ok, let's merge this branch, so that we have a decently documented master for now. I will just open another branch later, where everyone can commit cosmetics and documentation.

PR #57, opened 2019-05-13
RNO phased array

daniel-zid:
The PhasedArray branch was too old, so I created this copy with the new files. This branch contains the phased array trigger with an updated configuration. There are some examples in the NuRadioReco/examples/PhasedArray/ folder than can be used as a basis for a more complete study of the phased array.


daniel-zid:
I've updated the code with some documentation and changed the event generator for the effective volume example. Please tell me if the documentation is not clear and I'll modify it. The examples should work immediately with a complete NuRadioReco and NuRadioMC installation.

daniel-zid:
The noise_trigger_rate.py is actually quite important, as it allows to
study the noise
trigger rate and therefore the correct threshold. I don't mind moving it to
the examples
folder.

The results are just a sample of what one can expect for the SNR curves if
properly
executed. I can remove it.

anelles:
I also thought that the results were too much, but they are tiny and I have to say that I found them helpful. These examples need quite heavy lifting for the simulations so having some results to see what they will look like sounds good to me.

daniel-zid:
Ooops, I've just removed them.

daniel-zid:
Aaaand, they're back.

PR #59, opened 2019-05-16
Timing

christophwelling:
Changes to make sure times are handled consistently:
- The station time is stored as a astropy.time object. This introduces astropy as a new dependency. For backward compatibility, base_station.set_station_time will still accept datetime objects and convert them to astropy.time
- detector.py also saves its time as a astropy.time object. For backward compatibility, detector.update still accepts datetime as parameter.
- Added set_position and get_position to the electric_field class. Timings regarding the electric field are defined for this position.
- Made sure the E-field reconstructed by voltageToEfieldConverter and voltageToAnalyticEfieldConverter has the correct trace_start_time relative to the channels.
- Made the channelLengthAdjuster change the channel trace_start_time to keep the timing consistent.
- Trigger simulators no longer change the channel traces. Instead the triggerTimeAdjuster module was added to do this. This should make it clearer what is happening (it is not intuitively clear that adding a trigger should change the traces) and avoid code duplication.
- Eventbrowser updated to handle these changes


christophwelling:
One more thing: In the __init__ of detector.py, I removed the lines
`        # just for testing
        self.__current_time = datetime.now()`

I assumed that this meant someone forgot to remove this after testing. It is probably better to make the user set the detector time explicitly, but it may break some scripts.

christophwelling:
- Comments on E-field position get/set function are now docstrings
- Yes, the variable antenna_position is now the antenna position relative to the E-field position. In practice this has no effect since air shower E-fields should be at [0,0,0], but it can't hurt to future-proof this.
- Yes, the efieldToVoltageConverter makes all channels have the same trace_start_time. The hardwareResponseIncorporator changes the trace_start_time to adjust for cable delays when it is run the second time, but this is supposed to happen after trigger simulation. So as far as I can see, if the trace_start_times differ something has probably gone wrong. No idea why I didn't check for this in all trigger modules, I probably just forgot. Fixed now.
- Deprecated trigger modules are now removed
- triggerTimeAdjuster now by default picks the trigger with the smalles trigger_time if no trigger name is given. I'm not sure if I understand you correctly: Should the station_time also be changed by the trigger simulation? Also, this is going off-topic, but it might be a good idea to let the triggers store the IDs of the channels that triggered them.
- Yes, the voltageToEfieldConverter fits the E-field timing by correlating it with the channel traces. I originally wrote this for the efieldToAnalyticVoltageConverter, where I still think this is the best way to do it. I can look onto doing it analytically for the voltageToEfieldConverter, though.
- Yes, the fft is real, but the data type is still a complex number (whose imaginary part happens to be 0). Usually that doesn't matter, but sometimes it can cause an error.

cg-laser:
Great! Thanks for implementing the changes so quickly!

> * triggerTimeAdjuster now by default picks the trigger with the smalles trigger_time if no trigger name is given. I'm not sure if I understand you correctly: Should the station_time also be changed by the trigger simulation? Also, this is going off-topic, but it might be a good idea to let the triggers store the IDs of the channels that triggered them.
No I didn't mean to change the station time. Your implementation looks good now. 
And yes, it will be good to save the triggered channels too. And maybe even the individual trigger times per channel (but this is difficult to define, as multiple bins might caused the trigger). Yue asked me about this yesterday. I would say it is good to have but not the highest priority at the moment. 

> * Yes, the voltageToEfieldConverter fits the E-field timing by correlating it with the channel traces. I originally wrote this for the efieldToAnalyticVoltageConverter, where I still think this is the best way to do it. I can look onto doing it analytically for the voltageToEfieldConverter, though.

the problem with this method is that it depends on your sampling rate. You might improve that by using the peakutils gauss fit peak approximation: https://peakutils.readthedocs.io/en/latest/tutorial_a.html

But I still think that doing it analytically for the voltageToEfieldConverter would be better. 

> * Yes, the fft is real, but the data type is still a complex number (whose imaginary part happens to be 0). Usually that doesn't matter, but sometimes it can cause an error.

really? I thought it was real. Then we might want to change this directly in the fft utilities to make sure that a real valued array is returned for the irfft.

christophwelling:
I created an issue for the changes to the trigger. I also noted a problem with the trigger modules: It looks like the trigger time is set to the beginning of the coincidence window instead of the timing of the actual pulse. I created an issue for that too.

The efieldToVoltageConverter calculates the trace_start_time analytically now. 

Now that you say it, it is weird that the fft returns a complex number. When we use get_trace and get_frequency_spectrum, the base_trace ffts all the time and seems to alway return real traces. Otherwise I think we should have noticed already. I checked, but couldn't get the get_trace function to return me a complex trace. I have no idea why that is happening when I calculate the voltages from the E-field, but it seems to only happen there. Maybe some rounding errors?

PR #62, opened 2019-05-23
VEL issue

anelles:
When setting up the build bot for NuRadioMC, I noticed a somewhat dangerous behavior that if this module breaks out of the except statement, VEL is still set in the loop, hence, the reconstruction continuous with using the wrong antenna model. This patch sets VEL = None which should explicitly break it if this occurs.

### DIFF
--- a/NuRadioReco/modules/efieldToVoltageConverter.py
+++ b/NuRadioReco/modules/efieldToVoltageConverter.py
@@ -188,14 +188,14 @@ class efieldToVoltageConverter:
                 azimuth = electric_field[efp.azimuth]
 
                 # get antenna pattern for current channel
-                try:
-                    VEL = trace_utilities.get_efield_antenna_factor(sim_station, ff, [channel_id], det, zenith, azimuth, self.antenna_provider)[0]
-                except:
-                    logger.error("efieldToVoltageConverter cannot continue with unphysical values, VEL is None")
-                    VEL = None
+                VEL = trace_utilities.get_efield_antenna_factor(sim_station, ff, [channel_id], det, zenith, azimuth, self.antenna_provider)
 
-                # Apply antenna response to electric field
-                voltage_fft = np.sum(VEL * np.array([efield_fft[1], efield_fft[2]]), axis=0)
+                if VEL is None: # this can happen if there is not signal path to the antenna
+                    voltage_fft = np.zeros_like(efield_fft[1])  # set voltage trace to zeros
+                else:
+                    # Apply antenna response to electric field
+                    VEL = VEL[0] # we only requested the VEL for one channel, so selecting it
+                    voltage_fft = np.sum(VEL * np.array([efield_fft[1], efield_fft[2]]), axis=0)
 
                 # Remove DC offset
                 voltage_fft[np.where(ff < 5 * units.MHz)] = 0.
--- a/NuRadioReco/modules/efieldToVoltageConverter.py
+++ b/NuRadioReco/modules/efieldToVoltageConverter.py
@@ -191,7 +191,8 @@ class efieldToVoltageConverter:
                 try:
                     VEL = trace_utilities.get_efield_antenna_factor(sim_station, ff, [channel_id], det, zenith, azimuth, self.antenna_provider)[0]
                 except:
-                    logger.warning("efieldToVoltageConverter cannot continue with unphysical values, VEL is None")
+                    logger.error("efieldToVoltageConverter cannot continue with unphysical values, VEL is None")
+                    VEL = None
 
                 # Apply antenna response to electric field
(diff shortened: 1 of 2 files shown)


NuRadioReco/modules/efieldToVoltageConverter.py
```diff
@@ -191,7 +191,8 @@ def run(self, evt, station, det):
                 try:
                     VEL = trace_utilities.get_efield_antenna_factor(sim_station, ff, [channel_id], det, zenith, azimuth, self.antenna_provider)[0]
                 except:
-                    logger.warning("efieldToVoltageConverter cannot continue with unphysical values, VEL is None")
+                    logger.error("efieldToVoltageConverter cannot continue with unphysical values, VEL is None")
+                    VEL = None
```
christophwelling: How about just throwing an error here? The next line should crash anyway if VEL is none, and that would make it easier to understand where the problem is.

cg-laser: why is this in a try except block anyway? I would just remove it. The only thing I can think of is that the zenith angle refracted into the ice is not defined somehow.

anelles: I am open to other suggestions of how to handle this. I know that this is not particularly elegant. The get_efield_antenna_factor returns None for unphysical values so indexing into it with [0] breaks the code. That is why we have the try statement because otherwise you only get the non-descript error, couldn't index.

cg-laser: I checked the code and the try except statement does nothing. The function just returns None if no refraction out of the ice is possible. So this error should only happen for antennas above the firn, can you confirm this @anelles ?
In that case we should set the voltage trace to zeros. And I'm pretty sure I implemented it like this some time ago. But maybe this got lost with the new efield object consolidation. 
So just remove the try/except, then check if VEL is None and set voltage_fft to zeros.

anelles: I agree that the error that @cg-laser found is very rare. 
However, the error also occurs, if the code cannot find an antenna model. Then the indexing breaks. It happened when the built-bot continued in the reconstruction without having finished the download of the antenna models. I am doing this manually now, but I would like something in addition as a safeguard for future occurrences. 
One could add something in the antenna class, I guess. Ideas?

cg-laser: Can you post the error message? What you describe should also not happen. It is implemented that a warning is issued when an antenna model couldn't be found and just an array of ones is returned. 
And my other suggestion should also be implemented, otherwise station 51 simulations will crash.

anelles: On the build-bot the automatic antenna model downloading does not work. We get:
WARNING:antennapattern:antenna pattern /home/travis/build/nu-radio/NuRadioMC/NuRadioReco/NuRadioReco/detector/AntennaModels/createLPDA_100MHz_InfFirn/createLPDA_100MHz_InfFirn.pkl does not exist, file will be downloaded
WARNING:efieldToVoltageConverter:efieldToVoltageConverter cannot continue with unphysical values, VEL is None
Traceback (most recent call last):
  File "NuRadioMC/test/T02RunSimulation.py", line 96, in <module>
    sim.run()
  File "/home/travis/build/nu-radio/NuRadioMC/NuRadioMC/simulation/simulation2.py", line 387, in run
    self._detector_simulation()
  File "NuRadioMC/test/T02RunSimulation.py", line 34, in _detector_simulation
    efieldToVoltageConverter.run(self._evt, self._station, self._det)  # convolve efield with antenna pattern
  File "/home/travis/build/nu-radio/NuRadioMC/NuRadioReco/NuRadioReco/modules/efieldToVoltageConverter.py", line 197, in run
    voltage_fft = np.sum(VEL * np.array([efield_fft[1], efield_fft[2]]), axis=0)
UnboundLocalError: local variable 'VEL' referenced before assignment

anelles: So when not finding the antenna model, it crashes with a warning that has very little to do with what went wrong.

anelles: So there are two problems:
1) what to do, if download fails, and why do some systems not wait for the download to complete?
2) the undefined zenith. I am not sure when this occurs, so @cg-laser can you take a look?

cg-laser: I looked into it and fixed the second part. 
The problem is that we don't know the actual error because of the try/except block. In general, avoid try/except as much as possible as this just hides the true errors. 
I have no idea what is going wrong with the download and I can't reproduce the bug. @anelles can you try again with my recent commit?

anelles:
The download starts but the rest of the script doesn't wait for it to be complete. I don't get an error message either ... I just see that VEL is None. When I download the file manually, the error goes away.

PR #63, opened 2019-05-23
Cosmic ray energy reconstructor

christophwelling:
Adds the cosmicRayEnergyReconstructor that reconstructs the energy of cosmic rays. Works Moores Bay and (with lower accuracy) the south pole.

Other changes:
- Adds parameters for Xmax and electromagnetic shower energy to parameter storage
- Adds the option to have a 2nd order parameter in the analytic E-field. If none is given the analytic_efield utility works as before
- Lets the voltageToAnalyticEfieldConverter perform this 2nd order fit


christophwelling:
- I didn't think I had enough simulations for a AERA-parametrization, but if I combine the simulations for sea level and 1500m (which I think is fine because the LDFs look very similar) it works. The parametrization actually looks surprisingly good, considering that we have a different magnetic field and altitude. So parametrization for AERA is now also in the module.
- Added bandpass filter details
- I will add a reference to the paper once it is on arxiv.
- Logger init added
- Not sure if an exception isn't overkill here. There may be good reasons for a station to not have an efield (e.g. not passing some quality cut) and killing the entire reconstruction because of that seems like a bit of an overreaction.

cg-laser:
@anelles here I guess it was fine but in general: Can we wait with the merging until the reviewer had the chance to look at the changes?

christophwelling:
We merged because we wanted to include the module's github link in the paper.

PR #64, opened 2019-05-24
handle missing config tree entries gracefully

cg-laser:


cg-laser:
yes in principle this should not happen. this is likely a result of the snowshovel merging bug. Thanks for putting in the final statistics.

PR #68, opened 2019-05-27
Update examples

anelles:
This is not finished. Examples either don't work or produce outdated results. 
Please add as you see fit.


cg-laser:
who is feeling responsible for fixing the examples?

christophwelling:
I worked on this a bit. The standalone scripts are running without error now. I have a feeling that the SNR comparison script is using the wrong channel IDs, but I am not sure. Could whoever wrote that example check please?
What should we do with the SimpleReconstruction example? The voltageToEfieldConverter does not work correctly for neutrinos and I don't really know what to use for a neutrino reco example

cg-laser:
I wasn't aware that we had a neutrino example. I agree, a reconstruction does not really make sense here. But we could show how to read in properties such as traces and sim parameters. Maybe just a channel loop and some print statements?

Issue #69, opened 2019-05-27
Phased array does not work with Python2


christophwelling:
The phased array trigger simulator works with python3 but not python2. If I try to import it, I get 'module not found'. 
Can it be made to work with 2.7? It is used in some of the examples. Since we "officially" support 2.7 we should either fix the module of remove the examples for now.

anelles:
It was only a missing init and some old modules. All fixed in pull request.

PR #70, opened 2019-05-27
Fix phasedarray

anelles:
Hotfix addressing issue #69 
It has nothing to do with python 3, it was just missing the init and was using an old trigger module.


anelles:
The threshold factors were always overwritten. This is actually just a clean up to avoid confusion.

PR #72, opened 2019-05-28
handle non existing 'stops' correctly in snowshovel reader

cg-laser:
previously an empty event was returned when no stop information was present


cg-laser:
this pull request also fixes a bug when the coincidence window is larger than the trace length. Now the coincidence window is restricted to the trace length.

PR #74, opened 2019-05-31
setting up NuRadioReco for first release on PyPi

cg-laser:
configuration and publishing is done using flit (https://flit.readthedocs.io/en/latest/index.html)
all dependencies are specified in the config such that NuRadioReco installation is just a `pip install NuRadioReco`

currently setting version to 0.1.0 for testing.


anelles:
Sounds good and nice work. We should also add this to the documentation, of course.

PR #75, opened 2019-05-31
fix trigger time. fixes #61, set default pre trigger time to 55ns (AR…

### LINKED ISSUE #61: Trigger times not set correctly
It looks like the trigger simulation modules set the trigger time to the start of the coincidence window around the pulse instead of the pulse.

cg-laser:
…IANNA default)

### DIFF
--- a/NuRadioReco/modules/trigger/multiHighLowThreshold.py
+++ b/NuRadioReco/modules/trigger/multiHighLowThreshold.py
@@ -1,5 +1,6 @@
 from NuRadioReco.utilities import units
 from NuRadioReco.framework.parameters import stationParameters as stnp
+from NuRadioReco.framework.trigger import HighLowTrigger
 from NuRadioReco.modules.trigger.highLowThreshold import get_majority_logic
 import numpy as np
 import time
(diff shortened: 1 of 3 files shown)


christophwelling:
Don't the other trigger modules need the same fix?

cg-laser:
the simple trigger module already used the function from the high/low module. Fixed the multiHighLowTrigger module too.

NuRadioReco/modules/trigger/multiHighLowThreshold.py
```diff
@@ -1,21 +1,21 @@
 from NuRadioReco.utilities import units
 from NuRadioReco.framework.parameters import stationParameters as stnp
-from NuRadioReco.framework.trigger import HighLowTrigger
```
christophwelling: I'm pretty sure this import is needed. Later a HighLowTrigger object is created.

cg-laser: thanks for spotting this! I somehow thought it was an outdated import of an old trigger module... I put it back in

cg-laser:
@anelles given that the bug I introduced was not caught by the built bot, can we add all trigger module to the example?

Issue #77, opened 2019-05-31
traces in event browser not plotted correctly when trace start time is huge


cg-laser:
**Describe the issue**
The x-axis range is often not set correctly if the trace start time is huge (which is always the case for neutrino simulations).  See screenshot below. 
![Screen Shot 2019-05-31 at 3 05 59 PM](https://user-images.githubusercontent.com/31775679/58737281-c2153600-83b5-11e9-8506-b0b190a78cdd.png)

Also the text (mean, RMS) is not displayed correctly. And the L1 value needs to be calculated on the absolute value of the spectrum. 

**Expected behavior**
For large but same trace start times can we do the following. Subtract the trace start time from the trace so that the trace starts as '0'. This is what we would measure. The large trace start time is just an artifact of the simulation and the new 'feature' that the trace start times are saved (which is needed to establish the timing wrt to the simulated efields).
The trace start time can be displayed as a text in the figure.

Issue #78, opened 2019-05-31
automatically open file in eventbrowser


cg-laser:
**Describe the issue**
When the eventbrowser is started from the commandline, only a folder can be specified.

**Expected behavior**
When a file is specified, directly open the file. 

In addition, maybe we could automatically open a window showing the event browser? I think one can start a browser window without any addressbar, bookmarks etc. 

Can we make the eventbrowser an executable? And the executable is automatically installed such that it is in $PATH. Then users can just execute "NuRadioRecoBrowser" or "NRRBrowser" from the command line.

cg-laser:
its seems that there is a nice python interface to the browser: https://www.codevoila.com/post/45/open-web-browser-window-or-new-tab-with-python but it doesn't work for me :-(

christophwelling:
I fixed the first two issues on the issues_77_78 branch. To open the browser window, I used the interface you linked to (worked for me), so can you see if it works for you? I decided to make this an optional, so to automatically open a browser window type python index.py /path/to/file --open_window. I imagine having this always on can easily get annoying. It also required turning the debug off, otherwise 2 browser windows will be opened. As far as I can tell that is an issue with dash itself.

cg-laser:
great! Works for me. Not sure what I did wrong. 

One enhancement could be to just open a bare browser window. At least for me it opens a new tab in Chrome. But I think it is good enough for now.

PR #79, opened 2019-06-04
Issues 77 78

christophwelling:
Fixes issues #77 and #78 :

- Both folders and files can be passed via the commandline argument. If a file is specified, it will be opene automatically.
- The option --open_window can be added to the commandline to automatically open a browser window
- Debugging had to be turned off. If debug output is on, the script will be run twice and 2 browser windows will be opened. This seems to be a bug with dash itself.
- The port can be specified in the commandline with the --port command. Useful in case the old port wasn't closed while testing the eventbrowser
- If all channels have a trace_start_time larger than 1000ns, the minimum trace_start_time, rounded down to the next 1000 will be subtracted from the times in the trace plots. The axis label is adjusted to show this.
- This should also fix the issue with the labels, but I don't have files with huge trace_start_times. Can somebody check real quick?

About making the event display an executable: Sounds like a good idea, but I have no idea how to do this.

### DIFF
--- a/NuRadioReco/eventbrowser/NuRadioViewer
+++ b/NuRadioReco/eventbrowser/NuRadioViewer
@@ -1,0 +1,4 @@
+#!/bin/bash
+
+DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
+python "$DIR"/index.py --open-window "$@"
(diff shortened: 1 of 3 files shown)


cg-laser:
Changes look great. Thanks for fixing it so quickly @christophwelling . 
I renamed the open window option. I newer saw cmdargs with an underscore. I also added an executable. Now users just need to add the eventbrowser directory to their $PATH and it works with just "NuRadioViewer folder".

PR #89, opened 2019-06-20
Arareader

uzairlatif90:
We would like to add the AraReader to NuRadioReco for making people happy


anelles:
Looking forward to lots of commits on this one from @uzairlatif90 @toej93 and @sdekockere :)

uzairlatif90:
Hey @anelles ! So for putting in the details for the ARA detector I am assuming you would want me to make .json file similar to the one here: NuRadioReco/NuRadioReco/detector/ARIANNA/arianna_detector_db.json for ARA. Is that correct? I will start working on making that for ARA. I am just a bit confused right now on what coordinate system would you want to me to use. ARA has a local system for each of its station and then a global system which has its origin in one of the corners of the ARA (incomplete) hexagon. I can also find the coordinates with respect to the southpole. 

The other thing was the interpolation function to make sure the spacings between the samples are the same. AraRoot does have one function that does this. I will try to include that within the reader this week.

anelles:
Hi @uzairlatif90, the coordinate system is still somewhat implicit in this scheme since not all sites require the same system. So put the channel coordinates relative to 0 (center of the station) and choose some local cartesian system for the global station positions. 
@cg-laser: See, we don't need a definition, but users still wonder what they should adhere to. This is why we need to define something and write instruction on how to change this if nothing else.  

The interpolation is probably good to have. Now that I  am learning that the ARA design has a drifting sampling frequency. 

My mention was purely intended to keep the development going. ;)

uzairlatif90:
Ok I will put in the antenna coordinates w.r.t to the station coordinate system. This is slighlty unconventional system in the sense that x-axis points towards the direction of ice flow. The global station coordinate system is just a translation from the South Pole though. 

I can put in the antenna coordinates w.r.t to the station origin after doing the rotation so that if someone does want to go from local to global they just have to do a translation then.

NuRadioReco/examples/AraData/readARAexample.py
```diff
@@ -0,0 +1,19 @@
+import NuRadioReco.modules.io.araroot.readARAData
+import logging
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger('readARAexample')
+import matplotlib.pyplot as plt
+from NuRadioReco.utilities import units
+
+readARAData = NuRadioReco.modules.io.araroot.readARAData.readARAData()
+
+n=readARAData.begin("/home/uzair/Documents/AraRootStuff/event9129.root")
```
christophwelling: It is not a good idea to have global paths on the repository. This will probably only work on Uzair's computer. It is better to let the user define the filename via the commandline: https://docs.python.org/3.3/library/argparse.html. If you want you can include a small example file in the repository and have the default point to it (with a relative path)

anelles: I think when we did this, there was no 'draft' for pull requests yet. This is clearly not finished. 
Also, this won't work for anybody but Uzair, because you need a working version of ARAroot ;)

anelles: But in principle, you are correct!

christophwelling: Okay, I just thought it was done and Uzair was waiting for someone to approve :-)

uzairlatif90: Thanks Christoph! For pointing that out. I am aware of that. Will certainly do that. Right now I am trying to get the interpolator running AraFile reader. In ROOT I am required to include a header file for it. In Python I am trying to call it through shared library .so file. Have been unsuccessful at it till now. Hopefully something will work out this week.

anelles: @uzairlatif90: don't try to think ROOT so much. Scipy has an incredible wealth of interpolation methods. https://docs.scipy.org/doc/scipy/reference/interpolate.html
It might be much easier -- and better for the sustainability -- to use a python for the interpolation. We are happy to help, when it comes to these programming things.

anelles:
Hi @uzairlatif90, thanks for pushing this. I have gone through and added documentation and cleaned up some minor things. For the lack of an ARA library, I cannot check whether this works as expected. Could you please have a look? After, I am happy to approve this and we can push it to master.

PR #92, opened 2019-06-26
remove t0

iplaisier:
trace start time does not have to be added


christophwelling:
@cg-laser Do you understand why the noise window is set the way it is? We couldn't figure it out, but since there is a 141 hardcoded into it it looks like this is still ARIANNA-specific. If so, we should probably fix this while we're at it

cg-laser:
I don't know... this doesn't look familiar to me

PR #94, opened 2019-07-01
Fixes ARA hardwareResponseIncorporator

christophwelling:
Fixes ARA hardwareResponseIncorporator. This bug was introduced when the get_channels function was changed to return an iterator.


christophwelling:
Hmm... apparently the hardwareResponseIncorporator produces nans. This is going to need a bit more work

christophwelling:
Okay, the problem was that for some frequencies the gain was 0, which causes problems when you divide by the gain. That is fixed now

PR #97, opened 2019-07-05
fix sign error in r_p coefficient. 

cg-laser:
The calculation of the Fresnel reflection and transmission coefficients are based on https://en.wikipedia.org/wiki/Fresnel_equations#Amplitude_reflection_and_transmission_coefficients. For total internal reflection cases, the formulas need to be modified as the outgoing angle is undefined. This is done in the following PDF document. 

[Fresnel_reflection_coefficients.pdf](https://github.com/nu-radio/NuRadioReco/files/3364126/Fresnel_reflection_coefficients.pdf)

For TIR incident angles, the reflection coefficients have magnitude one and change phase dependent on the incident angle. The previous implementation had a -1 error in the r_p coefficient. 

I also changed the code to return 0 for transmission coefficients for TIR cases

### DIFF
--- a/NuRadioReco/utilities/geometryUtilities.py
+++ b/NuRadioReco/utilities/geometryUtilities.py
@@ -109,6 +109,8 @@ def get_fresnel_t_p(zenith_incoming, n_2=1.3, n_1=1.):
     of the incoming radiation."
     """
     zenith_outgoing = get_fresnel_angle(zenith_incoming, n_2, n_1)
+    if(zenith_outgoing is None):
+        return 0
     t = 2 * n_1 * np.cos(zenith_incoming) / (n_1 * np.cos(zenith_outgoing) + n_2 * np.cos(zenith_incoming))
     return t
 
@@ -124,6 +126,8 @@ def get_fresnel_t_s(zenith_incoming, n_2=1.3, n_1=1.):
     of the incoming radiation."
     """
     zenith_outgoing = get_fresnel_angle(zenith_incoming, n_2, n_1)
+    if(zenith_outgoing is None):
+        return 0
     t = 2 * n_1 * np.cos(zenith_incoming) / (n_1 * np.cos(zenith_incoming) + n_2 * np.cos(zenith_outgoing))
     return t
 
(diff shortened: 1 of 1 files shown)


cg-laser:
Actually this 'fix' is not compatible with our measurement. We measure a direct and reflected pulse. According to the old formula, the phase shift should be ~55deg, which matched exactly our measurement (taking the first direct pulse as template and multiplying it (the FFT) with exp(1j * 55deg).  The template labeled with (180deg) is the result for this branch. 
ch6_R1_avg_zoom2

So either the old implementation was correct (but I don't remember why I had the -1 difference in there and can't reproduce it at the moment), OR the phase shift needs to be calculated with 
`exp(-1j * phase)` instead of `exp(1j * phase)` but I wouldn't know why.

cg-laser:
ok reading wikipedia more carefully (https://en.wikipedia.org/wiki/Total_internal_reflection#Phase_shifts)
`
In equations (5), (7), (8), (10), and (11), we advance the phase by the angle ϕ if we replace ωt by ωt+ϕ  (that is, if we replace −ωt by −ωt−ϕ),  with the result that the (complex) field is multiplied by e−iϕ. So a phase advance is equivalent to multiplication by a complex constant with a negative argument. This becomes more obvious when (e.g.) the field (5) is factored as {\displaystyle \mathbf {E_{k}} e^{i\mathbf {k\cdot r} }e^{-i\omega t},\,} {\displaystyle \mathbf {E_{k}} e^{i\mathbf {k\cdot r} }e^{-i\omega t},\,} where the last factor contains the time-dependence
`
the new equation is correct but we need to multiply with exp(-1j phase). Which corresponds to multiplying with the negative reflection coefficient. And this seems to apply to all coefficients. So the best implementation would be to always return the negativ reflection/transmission coefficient (?) so that `E_reflected = E * r` and `E_transmitted = E * t`.

christophwelling:
The way it is implemented looks correct to me. I just don't understand what exactly you mean by "return the negative reflection/transmission coefficient"?

cg-laser:
do not merge, the current implementation is not correct

cg-laser:
> Please fix. I don't know how to put this back into a draft version.

If I knew the right answer, I would do it...

anelles:
What is the status of this? Given that the DnR paper is under review, I assume that @cg-laser has worked out the correct solution by now.

cg-laser:
with the DnR we are only testing the eTheta component which is correctly implemented in the master (but not in this pull request). For the ePhi component I'm not so sure

anelles:
This one is ancient. And we still don't have the correct solution. Should we move this on an issue instead and close this for now?

anelles:
@cg-laser you probably need to make the call on this one ...

cg-laser:
we won't merge this PR but I suggest to leave it open for a future reference on this discussion.

PR #98, opened 2019-07-08
mean option added

a-anker:
Created optional mean subtraction of trace. Set mean_opt=False to remove mean subtraction.


a-anker:
Checked new average deviation and compared to old signal dataset. New dataset increase average deviation (shift of trace) by a few mV.

PR #102, opened 2019-07-19
Add trig list

a-anker:
Added option to store all triggers of a given trace in an array.


a-anker:
Okay. It mentions ns in the set_trigger_time (not plural) as well. Is that okay?

cg-laser:
can you remove that as well?

Issue #52, opened 2019-05-03
Analytic pulse utilities needs documentation


anelles:
utilities/analytic_pulse.py has no documentation and the first function has no units, which makes it impossible to understand where 0.12876705 comes from. Please document.

PR #105, opened 2019-07-27
Documenting

anelles:
Ok, try again. 
Adding documentation, also addressing #52 
@cg-laser can you add some more to the analytic pulse.


anelles:
I think most commits are from @christophwelling by now, so I am taking the liberty of reviewing and merging.

PR #107, opened 2019-07-29
fixed argument name

fschlueter:
Hi Guys,
i had to fix this argument name to get it working (used python2 and followed the installation instruction for plotly dash from the wiki).

Sorry the spam if i missed something..


christophwelling:
You are correct, these arguments were removed in the latest dash version: https://github.com/plotly/dash/pull/753
This has probably broken more than just what you have fixed. I will take a look at it.

christophwelling:
Looks like there were only a handfull of things that needed fixing:
- Callback function take the argument 'value' instead of 'values'.
- Dash no longer loads missing stuff from the web by default. We could turn that back on, but I think it is better to just use the offline mode (so you can use it anywhere, even, let's say, in Antarctica). Therefore the fontawesome icons are now part of the repository.
- The warning for the minimum required dash version was updated to 1.0. (You will need to update dash now)

anelles:
I am not sure that anyone can review > 1600 files ....

cg-laser:
I think Felix unintentionally commited thousands of unneeded files. This should be removed.

cg-laser:
> I think Felix unintentionally commited thousands of unneeded files. This should be removed.

sorry these are actually "font-awesome" icons. @christophwelling do we really need all these files? I would try to avoid clogging our software with thousands of css files. But I agree, it should also work offline. But wasn't it already working offline before? I thought so.

christophwelling:
Yes, that was me including fontawesome icons. I didn't see that it included so many files. Sorry about that.
Looks like there are generators that can generate custom fonts that include only the icons you need, so I'm gonna use one of those to downsize this.
Dash used to load some stuff online by default. It's a setting that can be changed, but we never did. So if there was no internet connection, the browser window would just never finish loading. With the lates dash version the default setting has been changed, and as I said, I think that is a good change.

christophwelling:
Okay, I removed all the unnecessary files again

fschlueter:
I think this is now good to go!
Thanks to @christophwelling  to properly fix it!

PR #108, opened 2019-07-29
When using useMCdirection=True and not running the correlationDirecti…

fschlueter:
Hi, 
in some lines a reconstructed zenith angle is used regardless of the useMCdirection option. This lead to a KeyError when "run" a reconstruction with useMCdirection=True and without a direction reconstruction. 

In the front of the running function the zenith and azimuth are set according to the flag.

Cheers,
Felix


anelles:
@fschlueter : Just FYI, the NuRadioReco work flow is that someone approves the pull request and then the author of the pull request merges it, when they are sure that this is the right time. Just in case you were waiting for anything from us ...

fschlueter:
Ah, okay. Thanks for the hint :)

PR #109, opened 2019-08-02
Add function to read-in AERA style antenna pattern from xml files

fschlueter:
Hi guys,

as the title indicates i would like to use the aera pattern to test your forward folding. I worte a function to read-in the AERA/offline xml files. However the results I am getting are looking wrong (see attached picture):
Screenshot_20190802_092007
To get the attached results I run a sequence like this:
        efieldToVoltageConverter.run(evt, station, det)
        channelBandPassFilter.run(evt, station, det, passband=[25 * units.MHz, 100 * units.MHz], filter_type='butter', order = 10)
        channelStopFilter.run(evt, station, det)
        channelBandPassFilter.run(evt, station, det, passband=[30 * units.MHz, 80 * units.MHz], filter_type='rectangular')
        channelResampler.run(evt, station, det, sampling_rate=0.8 * units.GHz)

        # traditional
        voltageToEfieldConverter.run(evt, station, det, use_channels=used_channels_efield, use_MC_direction=True)
        electricFieldSignalReconstructor.run(evt, station, det)

I actually can not find the reason for that. Can you maybe give me a hint?
Cheers,
felix


christophwelling:
It looks like you are getting these artifacts from dividing by the antenna response, which is essentially 0 for those frequencies. This blows up any frequencies that are not exactly 0.
I have 2 ideas to fix this:
1. The voltageToEfieldConverter checks if the antenna response is 0 for any frequency and ignores those (line 161). If the AERA antenna response is not exactly 0, try changing this check to < some really small number.
2. You applied a rectangular filter, so that should have set the voltage in those frequency bins to 0 already. The resampler you run afterwards probably produces some rounding errors and makes these slightly larger than 0 again. Having the filter after the resampler might fix this.

2 other points:
- This fits better into an issue than a pull request.
- The voltageToAnalyticEfieldConverter assumes that the voltage trace has a butterworth filter applied to it, so if you want to use it, it's better to make the 25-100MHz filter the rectangular one and the 30-80MHz the butterworth. The impact this has on the result is not that large, but noticable.

fschlueter:
Hi, 
thanks a lot for your comments! I will check it!
You are right, I got lazy not finding the bottom to connect the issue with this branch...  
But wait you just do #109 ... stupid me.

fschlueter:
Already looks much better. Still it looks like something is wrong. Any idea what could go wrong to?
grafik

cg-laser:
it the first one the reconstructed efield? The channel trace looks good to me

fschlueter:
> it the first one the reconstructed efield? The channel trace looks good to me

Yep. Really do you expect this second bumb. When there is a problem with the pattern (which I also think it is) it should be already visible in the channel trace (the second bumb?!), right?

cg-laser:
I don't really know how the pattern looks like, therefore I suggested that you plot the pattern as debug. 
With these line you can directly access the antenna pattern:
```python
antenna = provider.load_antenna_pattern("createLPDA_100MHz_infinitefirn")
df = 1 * units.MHz
ff = np.arange(0 * units.MHz, 1000 * units.MHz, df)
VEL = antenna.get_antenna_response_vectorized(ff, np.deg2rad(0), np.deg2rad(0),
                                              np.deg2rad(0), 0, np.deg2rad(90), np.deg2rad(0))
fig, ax = plt.subplots(1, 1)
ax.plot(ff / units.MHz, np.abs(VEL['theta']), '.-', label='eTheta')
ax.plot(ff / units.MHz, np.abs(VEL['phi']), '.-', label='ePhi')
ax.set_xlabel("frequency [MHz]")
ax.set_ylabel("vector effective length [m]")
fig.tight_layout()
ax.legend()
```

cg-laser:
BTW: already your efield has this double bump. Is the the simulated efield? Or the reconstructed one?

fschlueter:
> I don't really know how the pattern looks like, therefore I suggested that you plot the pattern as debug.
> With these line you can directly access the antenna pattern:
> 
> ```python
> antenna = provider.load_antenna_pattern("createLPDA_100MHz_infinitefirn")
> df = 1 * units.MHz
> ff = np.arange(0 * units.MHz, 1000 * units.MHz, df)
> VEL = antenna.get_antenna_response_vectorized(ff, np.deg2rad(0), np.deg2rad(0),
>                                               np.deg2rad(0), 0, np.deg2rad(90), np.deg2rad(0))
> fig, ax = plt.subplots(1, 1)
> ax.plot(ff / units.MHz, np.abs(VEL['theta']), '.-', label='eTheta')
> ax.plot(ff / units.MHz, np.abs(VEL['phi']), '.-', label='ePhi')
> ax.set_xlabel("frequency [MHz]")
> ax.set_ylabel("vector effective length [m]")
> fig.tight_layout()
> ax.legend()
> ```

Ahh Thanks 

> BTW: already your efield has this double bump. Is the the simulated efield? Or the reconstructed one?

Reconstructed, I just figured that you also have a efield bandpass filter.

fschlueter:
Hi all,
i had to fix some issues in my reconstruction sequence (see issue #110 and similar things). The result now looks quite reasonable to me:
grafik

I also looked at the response directly (using Christians code): (zeinth = 45deg)
LPDA_AERA_measured_45_0
The feature we saw previously is actually in the pattern too (just at a different zenith angle): (zenith = 0deg)
LPDA_AERA_measured_0_0
I checked with a similar pattern and this does not shows this peak at 55 MHz. Therefore I assume that the read-in is correct. I will do the necessary changes soon:

> Regarding your implementation: NuRadioReco requires the antenna pattern to be in a pickle format (which is also much smaller in size than the xml file from Offline). So you would need to write a reader for the XML file and export it into a pickle file. Then the correct antenna pattern is identified by its unique name. The new `source` parameter should be removed. A good unique name would be `AERA_smallblackspider_LPDA` or something similar.
> 
> You already wrote the xml reader function. Now you just need to add another "write to pickle" function similar to `save_preprocessed_WIPLD`. You should need to anything in the antenna pattern class.
> Now that I'm looking at the code, the function `get_WIPLD_antenna_response` is badly named, as it returns any antenna pattern in the pickle format.

Have a good vacation time :)
Cheers

anelles:
@fschlueter Yes, antenna models are somewhat black magic and you see things that you never had thought would be there. 
It would be great -- if we may ask -- if you can also commit your testing scripts to the repository that future people who need to add an antenna model have an easier time. It is still fresh in your memory. Thanks!

fschlueter:
> @fschlueter Yes, antenna models are somewhat black magic and you see things that you never had thought would be there.
> It would be great -- if we may ask -- if you can also commit your testing scripts to the repository that future people who need to add an antenna model have an easier time. It is still fresh in your memory. Thanks!

Hi, 
do you mean my reconstruction sequence? Sure I can do.
Cheers

fschlueter:
Alright,
should I than delete / revert the commit adding the reconstruction sequence again?

PR #111, opened 2019-08-05
Fix base station particle type

### LINKED ISSUE #110: Default particle type in station class
Hi guys,
it might be a design decision but it caused some trouble for me, this is why i am reporting this. 
In the base station class the particle type is initialized with
        self._particle_type = 'nu'
this forces you to run the cosmicRayIdentifier. Otherwise you transform the signal as a cr station and re transform it as a neutrino station which has effects on the applied zenith angle.
I think a more safe solution would be to initialize the stations without any type and raise an error in the set_is_cosmic_ray / set_is_neutrino function when the type is not initialized.
Cheers

fschlueter:
Hi guys,
this is how i would fix #110 . If it is foreseen that their will be more particle types than 'cr; and 'nu' one might just want to check for " if not self._particle_type == "": ".
I have not ran any test if the current StandardReconstruction would fail... (Is this covered by the CI?)
Cheers


fschlueter:
> fail... (Is this covered by the CI?)
Ha apparently it is covered, nice.

fschlueter:
The new commit might fix the cause, But not sure if you are happy with that.

christophwelling:
I think if more cosmic-ray focused experiments want to use our software, this is a good change. Otherwise this might easily trip up some people and make them mad at us ;-)
This will require also having a module that can flag the station as a neutrino. Instead of adding an additional module for this, I would prefer to turn the cosmicRayIdentifier into an 'eventTypeIdentifier' or something that can be made to set the flag either way.

NuRadioReco/framework/base_station.py
```diff
@@ -174,18 +174,27 @@ def get_electric_fields_for_channels(self, channel_ids=None, ray_path_type=None)
...
     def is_neutrino(self):
+        if not self._particle_type == 'nu' and not self._particle_type == 'cr':
+            logger.error("Stations particle type was not properly set: \"{}\"".format(self._particle_type))
+            raise ValueError("Stations particle type was not properly set: \"{}\"".format(self._particle_type))
+
         return self._particle_type == 'nu'
-        
+
     def is_cosmic_ray(self):
+        if not self._particle_type == 'nu' and not self._particle_type == 'cr':
```
christophwelling: I think it would be better to just check for particle_type == '' instead here. We just want to make sure it was initialized, but there may be other particle types that make sense.

fschlueter: Agree, I can change that.

fschlueter:
>  I would prefer to turn the cosmicRayIdentifier into an 'eventTypeIdentifier' or something that can be made to set the flag either way.

Should I do that? Do you want to have this in the same PR?

christophwelling:
I can do it if you want, I'm fine either way. We should do it in the same pull request, otherwise the master will be broken for a while.

fschlueter:
> I can do it if you want, I'm fine either way. We should do it in the same pull request, otherwise the master will be broken for a while.

Then go a head :) Hm unfortunately I did checked out from PR #109

fschlueter:
Hey @christophwelling what do you think how should we proceed with the two first commits in this branch?

christophwelling:
I think once that other branch is merged this should fix itself automatically.

fschlueter:
right so we have to wait, okay

christophwelling:
Replaced the cosmicRayIdentifier with a eventTypeIdentifier. As far as I'm concerned this can be merged now.

christophwelling:
The coreasReader automatically sets the particle type of the sim_station, so any module that implements the detector response can find it there.
The idea behind not setting it in the station is that in principle identifying cosmic rays/neutrinos is part of the reconstruction. We haven't implemented it yet, so the eventTypeIdentifier is a placeholder for now, but eventually we will need something that does this.

fschlueter:
Hi,
@cg-laser not sure that I can follow, a `git checkout antennapattern.py` would just remove changes in the stashed area right? This are commited changes (I unfortunately checked out from a other than master branch). One could try to revert or remove this single commit. But I dont know if that creates some artifacts in the git log.
Cheers

cg-laser:
@christophwelling I got it. It makes sense to not set it on station level. 

@fschlueter yes, reverting a commit is more complicated... `git revert` is what you're looking for: https://www.atlassian.com/git/tutorials/undoing-changes/git-revert You need to specify the commit hash-id and the command will create a new commit that inverts the old one.

cg-laser:
ok I reverted the changes to the antenna pattern class. Can be merged now.

PR #112, opened 2019-08-06
fix for mc direction reconstruction #108

fschlueter:


fschlueter:
Fix is similar to the one from #108

Issue #80, opened 2019-06-05
test all trigger module in built bot


cg-laser:
See #75 , we should test all trigger modules in the built bot!
In general it would be good if all modules would be used/tested with the built bot

PR #115, opened 2019-08-19
Unit tests

christophwelling:
Adds unit tests.
I started with a test for the trigger modules to address #80 .

@uzairlatif90 I think it would be a good idea to add a unit test for ARA. That way we can make sure that we don't accidentally break any ARA stuff. Can you make one?


christophwelling:
Update:
- The examples are included in the unit test. It was always a bit annoying to have the examples break from time to time. Now we know who to blame ;-)
- The test of TinyReconstruction now compares all parameters of the station, sim_station, channels and E-fields to a reference. This reconstruction includes most modules, so this should notice if something breaks.

cg-laser:
great! Just a general comment: The tests use NuRadioMC but shouldn't NuRadioReco be independent of NuRadioMC? I'm not sure if this is the best solution. Could a unrelated change in NuRadioMC make the test fail?

anelles:
Nice work!
@cg-laser: I think the alternative to using NuRadioMC would be to use a pre-made file, but then you are only testing half of the models. I think it is ok for the built-bot.
@christophwelling: Regarding your suggestion for @uzairlatif90, I don't think it is needed right now. No one can reliably install ARAroot -- I am certainly not willing to try it on the built-bot. And the modules require it. So we won't be able to do automated testing anyway.

christophwelling:
Yes, the unit test needs NuRadioMC and can fail if something in NuRadioMC changes.
We could remove the simulation parts and store pre-made files, but this would have the disadvantage that changes to the event structure may make them unreadable an we have to recreate them after large changes (like switching to py3). This is doable, but may be a bit annoying.

cg-laser:
I'm just afraid that we get failed test because of unrelated errors in NuRadioMC. And then the tests are of limited help. But maybe this will never happen, so I'm ok with merging the changes. If this becomes a problem we can still solve it later.

christophwelling:
I'm going to merge it as it is so we don't get any issues when changing to py3. We can still remove the simulation later

cg-laser:
ok

PR #116, opened 2019-08-20
adding preprocess_HFSS

iplaisier:
Adding a converter for HFSS to the RadioReco format. 

Effective length calculation still needs to be verified.


cg-laser:
we normally split up the functionality into two functions. One that parses the data file and returns its parameters. And another function that calculates the effective length and saves it to a file. Can you make this change? This allows to use the antennapattern class to parse these data files which might be useful at some point.

iplaisier:
I changed it

anelles:
@iplaisier does this work now? According to your results, it should, right? Can you then mark this as ready for review?

PR #121, opened 2019-08-27
Makes stations store triggers and channels in an ordred dict

christophwelling:
Triggers and channels are stored in an ordered dict, to make unit tests easier and avoid errors.
E-Fields are stored in an array, so their ordering is conserved anyway.


christophwelling:
E-Fields are stored in an array, not a dict. Arrays should keep their ordering anyway.

PR #124, opened 2019-08-29
This large commit initiates / improves the support for multi-station …

fschlueter:
…/ large array air shower detection.

A new class / module is added which reads full coreas simulations, that means all stations and most parameters. Sofar only the SIM*_highlevel.hdf5 files from the latest CoREAS trunk are supported.
A new Shower class is introduced which can store shower parameter. Each shower should belong to one event and has a identifier such that one event can have multipe showers.
To accommodate new requierments to support large array air shower detection new parameters are added.
Also an example application to read a SIM*_highlevel.hdf5 file and give some output is added.

Hi Guys,
this is the first draft. It gives as something to start discussing about. 
You may can see the general idea I have. However a lot can / should be improved. For example it is probably better to store showers in the same sense like stations in an event...
Maybe you can tell me already if I violated some design choices. For example I added a, i think, useful member function to the SimStation class returning the axis distance of the station (if the position in the vxB vxvxB system is set). However this requires some imports at the top and I also have not seen similar functions in other classes. Also the naming of for example the parameters might need to be adjusted.
One other Idea I was thinking about is: We could create new classes for stations such as AERA and ARIANNA stations and set the __setitem__ __getitem__ to the corresponding parameter dictionaries.

What do you think about this?
Cheers,
Felix


cg-laser:
did you forget to commit the coreasShowerReader

christophwelling:
We have so far taken the magnetic field from the radiotools. Do you need to store the magnetic field directly? The you are right, the event parameters make sense. I just didn't think of that.

fschlueter:
Hi, 
i think in principle one want to have a time dependent magnetic field and store it per event (or one has this functionality in the detector).
However I think than one want to store with which magnetic field the event was reconstructed.

christophwelling:
I think the best way to do it is to have a get_sim_showers at the event level. We want to be able to have multiple sim showers for stuff like tau events and that wouldn't really work if the sim shower was part of the radio_shower.
The sim_shower wouldn't event need to be its own class. As far as I can see, all it needs to do is have a parameter storage.

cg-laser:
> Hi,
> i think in principle one want to have a time dependent magnetic field and store it per event (or one has this functionality in the detector).
> However I think than one want to store with which magnetic field the event was reconstructed.

we definitively need to store the magnetic field of a simulated event, and the sim_station has a getter/setter function to store it. 

Not sure if we need to store the magnetic field for a reconstructed station/event if we have the utility function to get it on the fly from the location and time.

NuRadioReco/framework/hybrid_shower.py
```diff
@@ -0,0 +1,16 @@
+import NuRadioReco.framework.base_shower
+
+class HybridShower(NuRadioReco.framework.base_shower):
```
fschlueter: Hi, 
when inheriting from the base_shower this means, that only showerParameters can be stored in the parameter storage. Is that really feasible to cover all parameter for possible hybrid shower? I am not sure about this. However otherwise it might just be to dirty to not do it the same way. What do you think?
Cheers

fschlueter:
Hi, I want to add an ADST reader as soon as I find some time. This would require an dependency to root (pyroot) as well as offline (if one want to use the auger radio parameter enums (for read and storage?)). What do think about this? No problem at all or do you maybe have a tip how one can include this dependencies elegantly.
Cheers

NuRadioReco/modules/io/coreas/coreas.py
```diff
@@ -113,3 +120,29 @@ def make_sim_station(station_id, corsika, observer, channel_ids,  weight=None):
...
+
+    zenith, azimuth, magnetic_field_vector = get_angles(corsika)
+    sim_shower.set_parameter(shp.zenith, zenith)
+    sim_shower.set_parameter(shp.azimuth, azimuth)
+    sim_shower.set_parameter(shp.magnetic_field_vector, magnetic_field_vector)
+
+    energy = corsika['inputs'].attrs["ERANGE"][0] * units.GeV
+    sim_shower.set_parameter(shp.energy, energy)
+    sim_shower.set_parameter(shp.shower_maximum, corsika['CoREAS'].attrs['DepthOfShowerMaximum'] * units.g / units.cm2)
+    sim_shower.set_parameter(shp.core, np.array([0, 0, 0]))
```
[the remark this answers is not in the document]
christophwelling: The CoREAS reader always puts the shower core at position [0,0,0] and adjusts the station position, so in this case, this works.

NuRadioReco/framework/hybrid_shower.py
christophwelling: I expect the overlap between the two shower types to be very large. If some of the porperties are only used for one shower type, that is fine.
The hybrid shower class is only meant to store general shower properties, anything experiment-specific is supposed to go into a hybridDetector class

NuRadioReco/modules/io/coreas/coreas.py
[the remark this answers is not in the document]
fschlueter: okay, that we need to store the height of the antenna plane to correctly calculate atmospheric related quantities.

christophwelling:
We already have the ARIANNA- and the ARA-reader, which both need ROOT and some experiment-specific software to work. If it's only in the reader, that is fine. 
Just remember that we can't include it in the unit tests and won't do any maintenance for it, so you need to watch out that it doesn't break.

NuRadioReco/modules/io/coreas/coreas.py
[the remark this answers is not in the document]
christophwelling: Just noticed that you want to use this for the shower reader as well. I think it's a good idea to do that and avoid code duplication. So, can you remove that part here and put into the readCoREAS?

NuRadioReco/modules/io/coreas/coreas.py
```diff
@@ -113,3 +120,34 @@ def make_sim_station(station_id, corsika, observer, channel_ids,  weight=None):
...
+    sim_shower.set_parameter(shp.shower_maximum, corsika['CoREAS'].attrs['DepthOfShowerMaximum'] * units.g / units.cm2)
+    sim_shower.set_parameter(shp.refractive_index_at_ground, corsika['CoREAS'].attrs["GroundLevelRefractiveIndex"])
+    sim_shower.set_parameter(shp.magnetic_field_rotation,
+                             corsika['CoREAS'].attrs["RotationAngleForMagfieldDeclination"] * units.degree)
+    sim_shower.set_parameter(shp.distance_shower_maximum_geometric,
+                             corsika['CoREAS'].attrs["DistanceOfShowerMaximum"] * units.cm)
+
+    sim_shower.set_parameter(shp.observation_level, corsika["inputs"].attrs["OBSLEV"] * units.cm)
+    sim_shower.set_parameter(shp.primary_particle, corsika["inputs"].attrs["PRMPAR"])
+    sim_shower.set_parameter(shp.atmospheric_model, corsika["inputs"].attrs["ATMOD"])
```
christophwelling: This property doesn't exist in out CoREAS files. Could you check for that before adding it?

fschlueter: In newly converted hdf5 files it should exists (I added it to the hdf5 converter in the CORSIKA/CoREAS trunk, and it should also be in the newest release). So for the moment it might be practical to check if it is in but i think for the long run we want this to be there in any case I think.

NuRadioReco/modules/io/coreas/coreas.py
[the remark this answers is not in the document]
fschlueter: the whole function? I can.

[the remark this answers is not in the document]
christophwelling: I meant just the parts that are different when you read the entire shower, for example the core

NuRadioReco/modules/io/coreas/coreas.py
christophwelling: Well, our files are a bit older. There is no harm in letting people read older files

NuRadioReco/modules/io/coreas/coreas.py
```diff
@@ -143,7 +143,8 @@ def make_sim_shower(corsika):
 
     sim_shower.set_parameter(shp.observation_level, corsika["inputs"].attrs["OBSLEV"] * units.cm)
     sim_shower.set_parameter(shp.primary_particle, corsika["inputs"].attrs["PRMPAR"])
-    sim_shower.set_parameter(shp.atmospheric_model, corsika["inputs"].attrs["ATMOD"])
+    if 'ATMOD' in corsika['inputs'].attrs.keys():
```
fschlueter: thanks

NuRadioReco/modules/io/coreas/coreas.py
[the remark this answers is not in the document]
fschlueter: Ah you meant using this function and just overwriting certain parameters? ind the readCoREASShower?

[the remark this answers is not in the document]
christophwelling: I mean you set all the parameters that are the same for both cases (which I think should be all except the core position) in coreas.py and the other ones in readCoREAS.py or readCoREASShower.py

NuRadioReco/framework/base_station.py
```diff
@@ -38,6 +58,8 @@ def get_parameter(self, key):
         if not isinstance(key, parameters.stationParameters):
             logger.error("parameter key needs to be of type NuRadioReco.framework.parameters.stationParameters")
             raise ValueError("parameter key needs to be of type NuRadioReco.framework.parameters.stationParameters")
```
cg-laser: I don't agree with this change. We still want to reconstruct these properties on station level, so we need to store them on station level! Otherwise we can't reconstruct an event with multiple stations anymore! Only modules that use all stations should write their reconstruction into the event or shower class. 

For simulated events this might be different, as we know the one true shower properties.

NuRadioReco/modules/correlationDirectionFitter.py
```diff
@@ -227,22 +228,24 @@ def ll_regular_station_fft(angles, corr_02_fft, corr_13_fft, sampling_rate, posi
...
             ax.set_ylabel("Correlation Ch 0/ Ch2", fontsize='small')
             plt.tight_layout()
 #             plt.close("all")
 
-        station[stnp.zenith] = max(ZenLim[0], min(ZenLim[1], ll[0][0]))
-        station[stnp.azimuth] = ll[0][1]
-        output_str = "reconstucted angles theta = {:.1f}, phi = {:.1f}".format(station[stnp.zenith] / units.deg, station[stnp.azimuth] / units.deg)
-        if station.has_sim_station():
+        shower = evt.get_first_shower([station.get_id()])
+        shower.set_parameter(shp.zenith, max(ZenLim[0], min(ZenLim[1], ll[0][0])))
```
cg-laser: this is wrong for neutrinos. The shower direction is not the incoming signal direction. The incoming signal direction should be saved in the station parameters (as before).

NuRadioReco/modules/voltageToEfieldConverter.py
```diff
@@ -137,14 +138,15 @@ def run(self, evt, station, det, debug=False, debug_plotpath=None, use_channels=
...
+        if use_MC_direction and evt.has_sim_shower():
+            sim_shower = list(evt.get_sim_showers())[0]
+            zenith = sim_shower[shp.zenith]
+            azimuth = sim_shower[shp.azimuth]
             sim_present = True
         else:
-            logger.info("Using reconstructed (or starting) angles as no signal arrival angles are present")
-            zenith = station[stnp.zenith]
-            azimuth = station[stnp.azimuth]
+            shower = evt.get_first_shower()
```
cg-laser: this also seems confusing. It seems that you can create several sim showers associated with the station. This will lead to not well defined efield reconstructions.

NuRadioReco/modules/io/coreas/readCoREASShower.py
```diff
@@ -0,0 +1,160 @@
...
+                names = obs_plane["antenna_names"]
+
+                # later parse station id from antenna name if possible
+                for idx, pos in enumerate(np.array(antenna_position)):
+
+                    station_id = antenna_id(names[idx], idx)  # return proper station id if possible
+                    station = NuRadioReco.framework.station.Station(station_id)
+
+                    sim_station = NuRadioReco.framework.sim_station.SimStation(station_id, position=pos)
+                    electric_field = NuRadioReco.framework.electric_field.ElectricField([0, 1, 2])
```
christophwelling: The channel IDs should not be hardcoded. Use the channel IDs from the detector description instead.

NuRadioReco/modules/io/coreas/readCoREASShower.py
```diff
@@ -0,0 +1,160 @@
...
+
+                    station.set_sim_station(sim_station)
+                    evt.set_station(station)
+
+            else:
+                for idx, (name, observer) in enumerate(f_coreas['observers'].items()):
+                    station_id = antenna_id(name, idx)  # return proper station id if possible
+
+                    station = NuRadioReco.framework.station.Station(station_id)
+                    sim_station = coreas.make_sim_station(station_id, corsika, observer, channel_ids=[0, 1, 2])
```
christophwelling: Same here

christophwelling:
Properties reconstructed by a single station are stored at station level again. It also should be fully backward compatible now (with any files produced after the py3 switch)

christophwelling:
@fschlueter What's the status with this? Is it ready for review or is there more work to be done?

fschlueter:
Hi, no I think you can review it.

NuRadioReco/examples/read_full_CoREAS_shower.py
```diff
@@ -0,0 +1,64 @@
...
+    print('Event {}'.format(event.get_id()))
+
+    # initialize detector
+    det = detector.GenericDetector(json_filename=args.detectordescription, default_station=102)
+    det.update(datetime.datetime(2011, 11, 11, 11, 11))
+
+    for station in event.get_stations():
+        station.set_station_time(datetime.datetime(2011, 11, 11))
+
+        sim_station = station.get_sim_station()
```
christophwelling: I think it would be better to get rid of the simStation.get_position method and instead add the new station directly in the coREAS reader. I created #156 to discuss this.

NuRadioReco/examples/read_full_CoREAS_shower.py
```diff
@@ -0,0 +1,64 @@
...
+                            'pos_altitude': position[2]}
+
+        det.add_generic_station(det_station_dict)
+
+    eventWriter.run(event, det)
+
+nevents = eventWriter.end()
+print("Finished processing, {} events".format(nevents))
+
+# import NuRadioReco.modules.io.eventReader
```
christophwelling: Please remove commented-out code

NuRadioReco/framework/parameters.py
```diff
@@ -62,5 +65,27 @@ class ARIANNAParameters(Enum):  # this class stores parameters specific to the A
...
+    zenith = 1  # zenith angle of the shower axis pointing towards xmax
+    azimuth = 2  # azimuth angle of the shower axis pointing towards xmax
+    core = 3  # position of the intersection between shower axis and an observer plane
+    energy = 4  # total energy of the primary particle
+    electromagnetic_energy = 5  # energy of the electromagnetic shower component
+    radiation_energy = 6  # totally emitted radiation energy
+    electromagnetic_radiation_energy = 7  # radiation energy originated from the electromagnetic emission
+    primary_particle = 8  # particle id of the primary particle
+    shower_maximum = 9  # position of shower maximum in slant depth, e.g., Xmax
+    distance_shower_maximum_geometric = 10  # distance to xmax in meter
```
christophwelling: How are these properly defined? Wouldn't that depend on the station position?

NuRadioReco/framework/sim_station.py
```diff
@@ -34,6 +35,14 @@ def get_simulation_weight(self):
...
         self.__simulation_weight = simulation_weight
 
+    def get_axis_distance(self):
+        if not self.has_parameter(astP.position_vB_vvB):
+            logger.error("sim station position in vB vvB v system is not set")
+            raise ValueError("sim station position in vB vvB v system is not set")
+
+        position_vB_vvB = self.get_parameter(astP.position_vB_vvB)
+        return np.sqrt(position_vB_vvB[0] ** 2 + position_vB_vvB[1] ** 2)
+
```
christophwelling: Isn't this deprecated?

NuRadioReco/modules/io/coreas/readCoREASShower.py
```diff
@@ -0,0 +1,109 @@
...
+            t_per_event = time.time()
+
+            filesize = os.path.getsize(self.__input_files[self.__current_input_file])
+            if(filesize < 18456 * 2):
+                logger.warning("file {} seems to be corrupt, skipping to next file".format(
+                       self.__input_files[self.__current_input_file]))
+                self.__current_input_file += 1
+                continue
+
+            if self.__verbose:
```
christophwelling: This stuff should be done using the logger

NuRadioReco/framework/parameters.py
fschlueter: Yes, depends. we would define it along the shower axis from xmax to the core (intersection of shower axis with an observer plane)

NuRadioReco/framework/sim_station.py
fschlueter: using the parameter it is deprecated. do you want to keep the functionality itself?

christophwelling: I prefer getting rid of it. I don't like the idea of mixing the detector description and the event.

fschlueter: okay

anelles:
@cg-laser, I think we need your review here, before things can move forward. I think @christophwelling is waiting for this to then tackle the EventBrowser again.

christophwelling:
Do we want to include #156 in this PR or merge now and fix that later?

NuRadioReco/modules/io/coreas/readCoREASShower.py
```diff
@@ -0,0 +1,137 @@
...
+            logger.info('Reading %s ...' % self.__input_files[self.__current_input_file])
+
+            corsika = h5py.File(self.__input_files[self.__current_input_file], "r")
+            logger.info("using coreas simulation {} with E={:2g} theta = {:.0f}".format(
+                self.__input_files[self.__current_input_file], corsika['inputs'].attrs["ERANGE"][0] * units.GeV,
+                corsika['inputs'].attrs["THETAP"][0]))
+
+            f_coreas = corsika["CoREAS"]
+
+            evt = NuRadioReco.framework.event.Event(corsika['inputs'].attrs['RUNNR'], corsika['inputs'].attrs['EVTNR'])
```
christophwelling: Is there any reason to not just iterate the event ids? The problem with the way it is right now is that for example in our simulations each event has the same id and run number, which causes problems. Are the ids in the CoREAS files actually important in practice?

fschlueter: Hi, they can be important. For a library with discrete bins in arrival direction and energy we used the run number as an identifier. 
The event number we do not use (but i would not be sure if that is the case for every one).

christophwelling: That could be an issue then. In the simulations we have, they all have the same ID.

fschlueter: One could have an additional option overwriting them, but this is somewhat messy

christophwelling:
Okay, the detector issue is solved now. The GenericDetector can hold a list of changes to the detector for a given event. The CoREAS shower reader adds the station positions to this list for each event. If you call the function set_event on the GenericDetector, it will give you the configuration for this event.
When the detector is stored in an event file, the changes list is also stored, so the information is still available. If you do NuRadioRecoio.get_detector(), it will give you the detector for the last event that it was asked for. If you get another event, the detector is updated automatically.
It's probably a good idea to put some instructions on how to use it in the documentation

NuRadioReco/modules/io/coreas/readCoREASShower.py
christophwelling: Maybe we could store it in the SimShower. That seems like a natural place to put information like this.
It's probably a good idea to have a place to put some more information about the simulation itself. That would also be useful for neutrinos, for stuff like which ice model was used

fschlueter: We should definitely store some more data on event level such as used atmospheric model, ...
I am still not sure if we should store something like the run_number on simshower lvl. I think this belongs to the event lvl.

christophwelling:
I left the SimStation.get_station_position in there to make it easier if everything works. I'm going to remove it before merging

cg-laser:
@christophwelling what happens if you don't read event sequentially. E.g. you first request event 1, then event 1000, and then 200.

christophwelling:
That works as well. Each of the methods in NuRadioRecoio that you could use to get an event automatically checks if there is a detector and updates it if necessary. So if you do NuRadioRecoio.get_detector() afterwards, it will always give you the right detector for the event.

NuRadioReco/modules/io/coreas/readCoREASShower.py
christophwelling: Yes, that is not ideal. But always taking the ID from the CoREAS causes problems as well. It should be at least an option to iterate the ID instead of taking them from the file.
Can you take care of that? I want to get this PR done.

christophwelling:
@cg-laser @fschlueter  Can we finally merge with or did you find any issues that need fixing?

cg-laser:
I'll go through all the code this afternoon and then give my ok. Then it should be ready to merge for you in the morning

NuRadioReco/detector/generic_detector.py
```diff
@@ -117,14 +138,77 @@ def add_generic_station(self, station_dict):
...
         if self.__default_station_id not in self._buffered_channels.keys():
             self._buffer(self.__default_station_id)
         self._buffered_channels[station_dict['station_id']] = {}
+
         for i_channel, channel in self._buffered_channels[self.__default_station_id].items():
             new_channel = copy.copy(channel)
             new_channel['station_id'] = station_dict['station_id']
             self._buffered_channels[station_dict['station_id']][channel['channel_id']] = new_channel
+
+    def add_station_properties_for_event(self, properties, station_id, run_number, event_id):
```
cg-laser: from the function definition it is not clear what 'properties' is. Can you add a standard doc string with more documentation? Is 'properties' a dictionary with the detector description parameters?

NuRadioReco/examples/CustomHybridDetector.py
```diff
@@ -0,0 +1,64 @@
...
+
+    #define some custom functions we want our detector to have
+    def set_data(self, data):
+        self.__data = data
+    def get_data(self):
+        return self.__data
+
+    #the custom detector must know how to serialize and deserialize itself
+    def serialize(self):
+        data_pkl = pickle.dumps(self.__data, protocol=2)
```
cg-laser: better use protocol 4 here.

NuRadioReco/examples/read_full_CoREAS_shower.py
cg-laser: Because we now handle positions in the detector, we should remove the following line. And probably also the `get_position` function from the sim station right?

NuRadioReco/framework/event.py
```diff
@@ -101,11 +103,51 @@ def get_stations(self):
     def set_station(self, station):
         self.__stations[station.get_id()] = station
 
+    def add_shower(self, shower):
```
cg-laser: please add documentation to these functions. In particular what data should be stored in a shower object. The same for the following functions.

NuRadioReco/modules/io/NuRadioRecoio.py
```diff
@@ -221,8 +242,26 @@ def get_detector(self):
         event is given.
         """
         if self._current_file_id not in self.__detectors.keys():
+            detector_dict = self._detector_dicts[self._current_file_id]
```
cg-laser: Can you add a few comments to the code saying what it being done here.

NuRadioReco/detector/generic_detector.py
```diff
@@ -51,46 +51,67 @@ def __init__(self, json_filename, default_station, default_channel=None, assume_
...
+
+    def _get_station(self, station_id):
+        if(station_id not in self._buffered_stations.keys()):
+            self._buffer(station_id)
+        res = copy.copy(self._buffered_stations[station_id])
+        if self.__run_number is not None and self.__event_id is not None:
+            for change in self.__station_changes_for_event:
+                if change['station_id'] == station_id and change['run_number'] == self.__run_number and change['event_id'] == self.__event_id:
+                    for name, value in change['properties'].items():
+                        res[name] = value
```
cg-laser: This implementation is slightly different than what I recall we discussed given that I understand it correctly. 

I thought changes to the detector description would be saved incrementally with each event. E.g. event 3 changes the antenna 1 to bicone, and event 6 changes antenna 2 to LPDA. Then event 7 would have both changes. 

seems to update a generic detector (the same for all events) with requested changes of this event. So the next event, it not containing its own overwrites, just return the generic detector. Or am I missing something? 

In any case, we should add more documentation to the code, maybe at the header of this file .

NuRadioReco/modules/io/NuRadioRecoio.py
```diff
@@ -196,13 +199,25 @@ def get_event_i(self, event_number):
         event.deserialize(evtstr)
         self.__read_lock = False
         self._current_file_id = file_id
+        self._current_event_id = event.get_id()
+        self._current_run_number = event.get_run_number()
+        if self._current_file_id in self.__detectors.keys():
+            if 'generic_detector' in self._detector_dicts[self._current_file_id]:
+                if self._detector_dicts[self._current_file_id]['generic_detector']:
+                    self.__detectors[self._current_file_id].set_event(self._current_run_number, self._current_event_id)
```
cg-laser: please add a comment describing the purpose of the lines above.

NuRadioReco/modules/io/coreas/readCoREASShower.py
cg-laser: This is not yet resolved right? I think we can't have events without a unique run/eventid pair.

NuRadioReco/modules/io/eventWriter.py
```diff
@@ -126,26 +136,40 @@ def __get_event_bytearray(self, event, mode):
...
+                    det.update(station.get_station_time())
+                    station_description = det.get_station(station.get_id())
+                    self.__stored_stations.append({
                     'station_id': station.get_id(),
                     'commission_time': station_description['commission_time'],
                     'decommission_time': station_description['decommission_time']
-                })
+                    })
+                else:
+                    station_description = det.get_raw_station(station.get_id())
```
cg-laser: I don't understand what a "raw_station" is. Please add an explanation to the doc string in generic detector.

NuRadioReco/detector/generic_detector.py
christophwelling: You are right, each change only applies to one event. I think doing it incrementally could cause problems if events are removed from the file. The detector is only saved for the events that are actually in the file. So in your example, if we applied some quality cuts and got rid of event 6, we wouldn't know that antenna 2 is an LPDA in event 7.
There are probably ways around this, but it seems pretty messy and error-prone, so I decided to store changes per event.

NuRadioReco/modules/io/coreas/readCoREASShower.py
fschlueter: Hi, 
I added a simple option to set a ascending run and event number. Are you fine with that?

christophwelling: Have you tested it? It looks weird. You use the parameter set_ascending_run_and_event_number in the run function, event though it is a parameter of the begin function.
Also, please explain what the parameter does in the docstring

NuRadioReco/detector/generic_detector.py
christophwelling: Done

NuRadioReco/modules/io/NuRadioRecoio.py
christophwelling: Done

christophwelling:
I addressed your requests. For the documentation, I think it is best to write an explanation on how to use the event structure and the detector description and put it into the documentation. The NuRadioReco paper is a bit outdated in some respects, so I think having a documentation that we can keep up to date is a better option. I started working on that on another branch.
Having two events with the same IDs can cause all kinds of trouble. If you do get_event, I'm not even sure which one you would get. The detector description also uses the event IDs, so that would also cause trouble here.
The changes are fully backward-compatible. I created a file on the master and opening it still works, even for the detector description.

NuRadioReco/detector/generic_detector.py
cg-laser: Sounds good, I actually prefer this implementation. So far this anyway only applies to coreas simulations where each event has a different station configuration.

NuRadioReco/modules/io/coreas/readCoREASShower.py
```diff
@@ -68,7 +70,12 @@ def run(self):
 
             f_coreas = corsika["CoREAS"]
 
-            evt = NuRadioReco.framework.event.Event(corsika['inputs'].attrs['RUNNR'], corsika['inputs'].attrs['EVTNR'])
+            if set_ascending_run_and_event_number:
+                evt = NuRadioReco.framework.event.Event(self.__ascending_run_and_event_number, self.__ascending_run_and_event_number)
+                self.__ascending_run_and_event_number += 1
+            else:
+                evt = NuRadioReco.framework.event.Event(corsika['inputs'].attrs['RUNNR'], corsika['inputs'].attrs['EVTNR'])
```
cg-laser: shouldn't we be robust here and check if we already have an event with the same id? Actually this should be checked by the eventWriter. But I think we should catch this. This might be a really difficult bug to find in the future. So better do it now properly.

cg-laser:
Awesome! The last open point that I can find is that the event writer should catch if an event with the same run/event id is written to the file.

NuRadioReco/modules/io/coreas/readCoREASShower.py
fschlueter: Hi,
have not tested it. I can try to do this next week. but it should be straight forward. I can also add it to the docstring.
Felix

fschlueter: Hi, 
puh that was very stupid.. sorry for that. It should work now. I tested it a bit and I hope it is fine now. The solution might be a bit unorthodox to use the value to set the run number and event id  also as bool but it saves one variable.
I also added some more documentation.
Cheers

christophwelling:
EventWriter now refuses to write events with the same ID and run number. I think we are finally done :-)

anelles:
I just approved the pull request, so @cg-laser you can merge and then finally this one can merge too. Too many people waiting for this.

cg-laser:
I already approved it yesterday and thought Christoph should have the honor to merge it ;-)

anelles:
Oh, most certainly. So @christophwelling whenever you are ready to push the button!

christophwelling:
Yay! :-)

PR #126, opened 2019-08-30
Beam forming new

ggaswint:
Added beamforming method for reconstructing direction.


ggaswint:
I don't understand why the extract times example is in there. I did not directly put that there today, I guess I don't understand github. I think my atom is forcing the spacing to be 8. Do I need to close this and make a new pull request?

ggaswint:
I think I have made the requested changes successfully.

cg-laser:
the doc string does not exactly describe what the function does. Doesn't it returns an array of the channel traces that is cut to the physical overlapping time?

Also plotting should not be part of the end method by default. But you can add this option for a debug mode.

PR #127, opened 2019-08-30
Add parameter

ggaswint:
Added internal clock to ARIANNA parameters, allows for millisecond timing accuracy.


ggaswint:
The internal clock keeps track of the time between triggers and records this in milliseconds. I believe it is just a readout of the internal clock. It is not a true time because it is reset after every trigger. I am not sure how the event_time is read, but I know if uses a different clock that doesn't get reset.

The example was from May and I am not sure how it is part of this push. I reverted it.

cg-laser:
before merging, please add the changes to the changelog.txt file

PR #132, opened 2019-09-04
add module decorator to keep track of applied modules

cg-laser:
First step towards keeping track of modules using decorators (see https://www.python-course.eu/python3_decorators.php for an introduction). 

All modules are saved 'per event' in the event class. We not only need to save the name of the module but also the instance because multiple instances of the same module can be used that differ in global parameters. Not sure how to keep track of the parameters of the begin or __init__ method because those functions don't have the event as parameters. Shall we introduce an additional global layer? Some class that keeps track of things that are the same for all events? Similar to the detector object? 

So far it allows to keep track of the execution time too.


cg-laser:
@christophwelling @anelles can we increase the numerical precision of the test? The only thing that changed was the multiplication order of the filters which might lead to very small deviations. The differences are at the relative 1e-6 level or even better.

cg-laser:
There is a problem with the current implementation when multiple stations are simulated/reconstructed. For every station the list of modules is extended... I guess we need to discuss a general design feature: Do we want to allow different module sequences for different stations? If yes, we would need to save the module sequence on station level which is also possible.

NuRadioReco/modules/base/module.py
```diff
@@ -0,0 +1,76 @@
...
+        print("starting timer")
+        import datetime                 
+        before = datetime.datetime.now()                     
+        x = original_function(*args,**kwargs)                
+        after = datetime.datetime.now()                      
+        print("Elapsed Time = {0}".format(after-before))
+        return x                                             
+    return new_function  
+
+def ModuleDecorator(Cls):
```
anelles: Do you actually need this? So far you have only used the run_decorator. If you want to use it, it should adhere a bit more to our coding standards. No print statements, NewCls should be changed in naming, etc.

NuRadioReco/modules/ARIANNA/hardwareResponseIncorporator.py
```diff
@@ -26,6 +27,33 @@ def begin(self, debug=False):
...
+            
+        if sim_to_data:
+            return amp_response * cable_response
+        else:
+            return 1./ (amp_response * cable_response)
+        
+    @run_decorator   
     def run(self, evt, station, det, sim_to_data=False, phase_only=False,mode=None,mingainlin=None):
         """
         Switch sim_to_data to go from simulation to data or otherwise.
```
anelles: Documentation should follow standards -- while at it.

NuRadioReco/modules/base/module.py
```diff
@@ -0,0 +1,76 @@
...
+
+def time_this(original_function):      
+    print("decorating")
+    def new_function(*args,**kwargs):
+        print("starting timer")
+        import datetime                 
+        before = datetime.datetime.now()                     
+        x = original_function(*args,**kwargs)                
+        after = datetime.datetime.now()                      
+        print("Elapsed Time = {0}".format(after-before))
```
anelles: This also looks like it is still in debug mode.

NuRadioReco/modules/channelBandPassFilter.py
```diff
@@ -7,7 +8,7 @@
 from NuRadioReco.detector import filterresponse
 import NuRadioReco.framework.sim_station
 
-
+# @ModuleDecorator
```
anelles: Remove, if not needed.

christophwelling:
I am in favor of storing the module sequence in the station. One of the features we are advertising is that you can be very flexible in which modules you run on which events and which stations, so we should try to keep this flexibility.

anelles:
I think I replied at the NuRadioMC pull that I would also advocate for station level parameters. Concerning the precision: @cg-laser: Isn't that related to the discussion we are having in 112 NuRadioMC ?

cg-laser:
> Are these all modules that are currently in use? 10 don't feel like many.

I was just lazy, for testing I just added a few.

NuRadioReco/modules/base/module.py
cg-laser: No it is not used, although it would be more elegant to make a class decorator for the full module, but I didn't get it to work...

cg-laser:
Ok then I'll implement it station by station. For modules that work on all events, I'll add another run_event decorator which then stores stuff on event level.

cg-laser:
I implemented the requested changes. 
There is still some open points:
- how to store __init__ and begin arguments? 
The begin function is actually annoying, we should merge __init__ and begin().

anelles:
I have been confused about init and begin every since. So I am all for merging ... merge it to init(), I would suggest

cg-laser:
@anelles @christophwelling I changed the numerical precision to 1e-6 for the validation. I think 1e-6 is good enough for all our calculations there. I also made the comparison recursive so that the complex dictionaries can be compared easily

NuRadioReco/modules/ARIANNA/hardwareResponseIncorporator.py
anelles: There is still no documentation for the function get_filter

NuRadioReco/modules/base/module.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def register_run(level=None):
+    if level not in ["station", "event"]:
+        raise NotImplementedError("The level needs to be either 'station' or 'event'")
+
+    def run_decorator(run):
+
+        @wraps(run)
+        def register_run_method(self, *args, **kwargs):
```
anelles: Please document these. They are hard to understand for those, who were not involved in making these. If we want to keep it sustainable, we need to document our choices.

cg-laser:
@anelles Thanks for being to annoying. Is it ok now?

NuRadioReco/framework/event.py
```diff
@@ -15,8 +16,45 @@ def __init__(self, run_number, event_id):
...
         self._id = event_id
-        self.__stations = {}
+        self.__stations = collections.OrderedDict()
         self.__event_time = 0
+        self.__modules_event = []  # saves which modules were executed with what parameters on event level
+        self.__modules_station = {}  # saves which modules were executed with what parameters on station level
+
+    def register_module_event(self, instance, name, kwargs):
+        """
+        registers modules applied to this event
```
anelles: There is documentation now, but that the register_module_event registers the modules applied to an event is kind of obvious :) Can you describe the parameter please?

NuRadioReco/modules/base/module.py
```diff
@@ -0,0 +1,79 @@
...
+            start = timer()
+            res = run(self, *args, **kwargs)
+            if(level == "event"):
+                evt.register_module_event(self, self.__class__.__name__, kwargs)
+            elif(level == "station"):
+                evt.register_module_station(station.get_id(), self, self.__class__.__name__, kwargs)
+            elif(level == "reader"):
+                # not sure what to do... function returns generator, not sure how to access the event...
+                pass
+#                 print(type(res))
```
anelles: Could you clean this up. There is a lot of unused/debug stuff in there.

NuRadioReco/modules/channelBandPassFilter.py
anelles: No?

NuRadioReco/framework/event.py
anelles: Sorry, I didn't mark these earlier.

cg-laser: fixed

NuRadioReco/modules/base/module.py
cg-laser: fixed

PR #141, opened 2019-09-09
Added a few fixes to make new software python3 compatable

ggaswint:
Added changes to import statement, and datetime function to make them python3 compatible.


cg-laser:
@ggaswint can you check that the times are reasonable? Maybe just print them for one event to see if the date is in the right ballpark. I'm not sure if the GetSec() function returns a unix timestamp.

ggaswint:
The dates and times are correct :)

PR #143, opened 2019-09-11
Writing efields

anelles:
Remove the efield stuff from channel. It is not clear to me, when whether there is actually a difference between mini and micro now ...

### DIFF
--- a/NuRadioReco/framework/channel.py
+++ b/NuRadioReco/framework/channel.py
@@ -55,9 +55,6 @@ class Channel(NuRadioReco.framework.base_trace.BaseTrace):
         data = {'parameters': NuRadioReco.framework.parameter_serialization.serialize(self._parameters),
                 'id': self.get_id(),
                 'base_trace': base_trace_pkl}
-        if(self.__electric_field is not None):
-            if(not(mode == 'micro' or mode == 'mini')):
-                data['electric_field'] = self.__electric_field.serialize()
         return pickle.dumps(data, protocol=2)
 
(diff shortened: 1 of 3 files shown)


anelles:
We need to discuss what to do about file size and make mini, micro and full better distinct. 250 KB per event without efields is huge for an array of 100 stations.

NuRadioReco/framework/channel.py
```diff
@@ -55,9 +55,6 @@ def serialize(self, mode):
         data = {'parameters': NuRadioReco.framework.parameter_serialization.serialize(self._parameters),
                 'id': self.get_id(),
                 'base_trace': base_trace_pkl}
-        if(self.__electric_field is not None):
-            if(not(mode == 'micro' or mode == 'mini')):
-                data['electric_field'] = self.__electric_field.serialize()
```
cg-laser: we should remove the `self.__electric_field` from everywhere

anelles:
Right now, it looks like 'full' stores everything, 'mini' only the efield and 'micro' only the channel. I find this backwards, I think the efields are least important. Any thoughts on this?

cg-laser:
for our cases we need more modes. Once we have a neutrino efield reco, the reconstructed efields are more important than the voltage traces. But we often don't need the simulated efields... Can we come up with a better scheme to select individual components? Maybe just arguments such as `save_sim_efield=True`, `save_voltage_traces=True`, `save_efields=True`

But we should do this in a fresh pull request, otherwise we might get merge conflicts with other changes to the framwork classes

Issue #138, opened 2019-09-09
Implement slot-in for hybrid detector


christophwelling:
To support generic information about hybrid detectors, the hybrid_shower needs to be able to handle generic hybrid_detector objects, as long as they fulfill some requirements(e.g. have a serialize/deserialize function).

PR #146, opened 2019-09-18
Handling the case when there are no electric fields

daniel-zid:
The code could not handle missing electric fields. This has been corrected.


daniel-zid:
@christophwelling Can you change that, please?

NuRadioReco/modules/efieldToVoltageConverter.py
```diff
@@ -222,10 +224,11 @@ def run(self, evt, station, det):
                 plt.show()
             if trace_object is None:  # this happens if don't have any efield for this channel
                 # set the trace to zeros
-                channel.set_trace(np.zeros(trace_length_samples), 1. / time_resolution)
```
cg-laser: I think this change and the two changes below are not required. This should put it back into the expected functionality. 
The bug was that the first channel that was considered didn't have any efield. But I think we can exclude that we have an event with no efields at all. So at least one efield is present such that  `trace_lenght_samples` and `time_resolution` and `times_min` are set

daniel-zid:
But it fails when I try to run a simulation. Why does it fail, if it's not
required?

cg-laser:
only lines 255ff shouldn't be required.

daniel-zid:
255ff?

daniel-zid:
If I remove those files, the problem persists. After writing them, the code began to run.

cg-laser:
the changes to lines 255 and following shouldn't be required

cg-laser:
@daniel-zid can you post the details of your simulation? Und what circumstances do you encounter this bug

daniel-zid:
Well, Christoph is witness. That's not the behaviour I encountered. I think
he has more insight on the topic than I do.

O 17:49, mér., 18/09/2019, Christian Glaser <notifications@github.com>
escribiu:

daniel-zid:
I's just a regular NuRadioMC simulation. With several stations. Maybe
that's the problem?

O 17:50, mér., 18/09/2019, Christian Glaser <notifications@github.com>
escribiu:

cg-laser:
Its not that I don't believe you. But you didn't fix the actual cause of the problem. And I need more information to be able to debug it. Can you post the NuRadioMC config, detector description etc. ?

daniel-zid:
Sure! This is the config:

signal:
  model: Alvarez2009
propagation:
  ice_model: greenland_simple
  attenuation_model: GL1
noise: False

And the detector description is just a single channel and multiple stations having the same channel. I am posting here the cropped detector file to keep it readable:

{
    "_default": {},
    "channels": {
        "1": {
            "adc_id": null,
            "adc_n_samples": 256,
            "adc_nbits": null,
            "adc_sampling_frequency": 1.0,
            "adc_time_delay": null,
            "amp_reference_measurement": null,
            "amp_type": "300",
            "ant_comment": "South Pole Station channel0",
            "ant_orientation_phi": 0.0,
            "ant_orientation_theta": 0.0,
            "ant_position_x": 0.0,
            "ant_position_y": 0.0,
            "ant_position_z": -100.0,
            "ant_rotation_phi": 90.0,
            "ant_rotation_theta": 90.0,
            "ant_type": "bicone_v8",
            "cab_id": "17-09",
            "cab_length": 5.0,
            "cab_reference_measurement": null,
            "cab_time_delay": 19.8,
            "cab_type": "LMR_400",
            "channel_id": 0,
            "commission_time": "{TinyDate}:2017-11-01T00:00:00",
            "decommission_time": "{TinyDate}:2038-01-01T00:00:00",
            "station_id": 101
        }
    },
    "comment": {
        "1": "a triggering array made of 8 vertical dipoles with 1GHz sampling, i.e., the bandwidth is 0-500MHz. A similar triggering array is found in the RNO proposal."
    },
    "positions": {},
    "stations": {
        "1": {
            "MAC_address": "0002F7F2E7B9",
            "MBED_type": "v1",
            "board_number": 203,
            "commission_time": "{TinyDate}:2017-11-04T00:00:00",
            "decommission_time": "{TinyDate}:2038-01-01T00:00:00",
            "pos_altitude": 0.0,
            "pos_easting": -5000.0,
            "pos_measurement_time": null,
            "pos_northing": -5000.0,
            "pos_position": "SP1",
            "pos_site": "southpole",
            "position": "SP1",
            "station_id": 101,
            "station_type": null
        },
        "10": {
            "pos_easting": -3571.4285714285716,
            "pos_northing": -3571.4285714285716,
            "station_id": 110
        },
ETC ETC
}

The steering file is a T02 quite close to the one used for the effective volume example. I convert to voltage, filter, and then apply a simple threshold at 1.5 sigma for each station.

cg-laser:
not needed. https://github.com/nu-radio/NuRadioMC/pull/138 fixes this problem

PR #150, opened 2019-09-20
New module antennadedispersion

cg-laser:
Adds a new module that dedisperses antennas.

Also fixed a typo in the detector module in the function `get_antanna_orientation`

### DIFF
--- a/NuRadioReco/modules/channelAntennaDedispersion.py
+++ b/NuRadioReco/modules/channelAntennaDedispersion.py
@@ -1,0 +1,79 @@
+from NuRadioReco.modules.base.module import register_run
+import numpy as np
+import copy
+import logging
+from NuRadioReco.utilities import units
+from scipy import signal
+from NuRadioReco.detector.antennapattern import AntennaPatternProvider
+from functools import lru_cache
+
+logger = logging.getLogger('channelAntennaDedispersion')
+
+
+class channelAntennaDedispersion:
+    """
+    
+    """
+
+    def __init__(self):
+
+        self._provider = AntennaPatternProvider()
+
+#     @lru_cache(maxsize=32)  # hashing doen't make sense because ff array is always different
+    def _get_response(self, det, station_id, channel_id, ff):
+        antenna_name = det.get_antenna_model(station_id, channel_id)
+        antenna = self._provider.load_antenna_pattern(antenna_name)
+        zen_ori, az_ori, zen_rot, az_rot = det.get_antenna_orientation(station_id, channel_id)
+        zen, az = None, None
+        if("LPDA" in antenna_name):
+            zen = zen_ori
+            az = az_ori
+        elif("bicone" in antenna_name or "dipole" in antenna_name):
+            zen = 90 * units.deg + zen_ori
+            az = 0
+        else:
+            raise AttributeError(f"antenna name {antenna_name} can't be interpreted")
+        print(zen, az, zen_ori, az_ori, zen_rot, az_rot)
+        VEL = antenna.get_antenna_response_vectorized(ff, zen, az, zen_ori, az_ori, zen_rot, az_rot)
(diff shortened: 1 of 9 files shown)


NuRadioReco/modules/channelAntennaDedispersion.py
```diff
@@ -0,0 +1,77 @@
...
+    def _get_response(self, det, station_id, channel_id, ff):
+        antenna_name = det.get_antenna_model(station_id, channel_id)
+        antenna = self._provider.load_antenna_pattern(antenna_name)
+        zen_ori, az_ori, zen_rot, az_rot = det.get_antenna_orientation(station_id, channel_id)
+        zen, az = None, None
+        if("LPDA" in antenna_name):
+            zen = zen_ori
+            az = az_ori
+        elif("bicone" in antenna_name or "dipole" in antenna_name):
+            zen = 90 * units.deg + zen_ori
```
christophwelling: Is it documented somewhere that LPDA and bicone have their orientation defined differently?

NuRadioReco/modules/channelAntennaDedispersion.py
```diff
@@ -0,0 +1,77 @@
...
+    def run(self, evt, station, det):
+        """
+        parameters
+        ----------
+        """
+        for channel in station.iter_channels():
+            ff = channel.get_frequencies()
+            response = self._get_response(det, station.get_id(), channel.get_id(), tuple(ff))
+
+#             trace = channel.get_trace()
```
christophwelling: Please remove this if it is not needed

NuRadioReco/modules/channelAntennaDedispersion.py
anelles: We should really make this consistently, or issue some warnings somewhere. I was not aware that this works like this.

NuRadioReco/modules/channelAntennaDedispersion.py
```diff
@@ -0,0 +1,77 @@
...
+        polarization = "phi"
+        if(np.sum(np.abs(VEL['theta'])) > np.sum(np.abs(VEL['phi']))):
+            polarization = "theta"
+        response = np.exp(1j * np.angle(VEL[polarization]))
+#         fig, ax = plt.subplots(1, 1)
+#         ax.plot(ff / units.MHz, np.abs(VEL['theta']), label='theta')
+#         ax.plot(ff / units.MHz, np.abs(VEL['phi']), label='phi')
+#         ax.set_title(f"{antenna_name} - {polarization}")
+#         ax.legend()
+#         plt.show()
```
anelles: Clean up, please.

christophwelling:
It's a typo fix. antAnna -> antEnna

NuRadioReco/modules/channelAntennaDedispersion.py
cg-laser: The orientation is not defined differently. The first vector is the orientation/direction of the antenna. But the sensitive direction for LPDAs and dipoles is different (by exactly 90 degrees). The LPDA is most sensitive when the signal comes from the boresight direction. A dipole is most sensitive when the signal comes in perpendicular to the antenna orientation (thus +90 deg + antenna orientation zenith angle).

cg-laser:
oki doki, I cleaned up the module. 
I also added a check to the efieldToVoltage converter to check if a station has any efields.

anelles:
Thanks, can you update the docstring in get_antenna_orientation ? It now talks about connectors. If we add here, what we mean for dipole and LPDA, I think this would alleviate some confusion.

cg-laser:
> Thanks, can you update the docstring in get_antenna_orientation ? It now talks about connectors. If we add here, what we mean for dipole and LPDA, I think this would alleviate some confusion.

@anelles done

Issue #160, opened 2019-11-11
Different channel trace start times


christophwelling:
Currently, some modules (for example the trigger modules) assume that  all channels have the same trace start time. This will not necessarily be the case, for example if we do the double trigger to catch both DnR pulses.

We should modify the modules in question to be able to handle different trace_start_times

cg-laser:
This is a tricky question. So far our definition of trace start time is that it is the same for all channels when the signal is being recorded. The `efieldToVoltageConverter` sets all channels to the same trace start time. And also all the remaining module executed before the `triggerSimulator` don't change the trace start time. This makes sense because the data that we get from a detector does not have any trace start time information. This is part of the reconstruction process to correct for delays that are different between channels. 

Also current trigger realizations (ARA, ARIANNA) don't allow for time offsets between channels when calculating the channel coincidences (although this would probably be possible as this happens within the FPGA). 

I don't think allow multiple traces per channels is the right way to go (see #162 ). 

I rather think if we want a trigger with different time offsets for the coincidence logic between channels, this should be an argument passed directly to the trigger module. Then it is the same as it would be implemented in hardware. 

BTW: If you have really long cables, you need to make sure that the prepulse time is long enough to contain the delay from the cables. In the forward simulation of the hardwareresponse, the traces are shifted in time according to the cable delays!

PR #161, opened 2019-11-12
Fix ara trigger

daniel-zid:
While the tunnel diode implementation seems to work just fine, the diode trigger for each channel was later ignored when calculating the coincidences. I've implemented a simpler method for calculating the coincidences - the code takes the minimum of the diode-filtered trace and stores the corresponding time as the trigger time. Then, using a sliding window, it checks if a requested number of coincidences lies in that window. I've also added a new method for the trigger class that calculates the threshold parameters power_mean and power_std using the bandwidth and noise RMS as input.

### DIFF
--- a/NuRadioReco/modules/ARA/triggerSimulator.py
+++ b/NuRadioReco/modules/ARA/triggerSimulator.py
@@ -113,6 +113,52 @@ class triggerSimulator:
 
         return trace_after_tunnel_diode
 
+    def calculate_noise_parameters(self,
+                                   sampling_rate = 1 * units.GHz,
+                                   min_freq = 50 * units.MHz,
+                                   max_freq = 1 * units.GHz,
+                                   amplitude = 10 * units.microvolt,
+                                   type='rayleigh'):
+        """
+        Calculates the mean and the standard deviation for the diode-filtered noise.
+
+        Parameters
+        ----------
+        sampling_rate: float
+            Sampling rate
+        min_freq: float
+            Minimum frequency of the bandwidth
+        max_freq: float
+            Maximum frequency of the bandwidth
+        amplitude: float
+            Voltage amplitude (RMS) for the noise
+        type: string
+            Noise type
+
+        Returns
+        -------
+        power_mean: float
+            Mean of the diode-filtered noise
+        power_std: float
+            Standard deviation of the diode-filtered noise
+        """
+        noise = NuRadioReco.framework.channel.Channel(0)
+
+        long_noise = channelGenericNoiseAdder().bandlimited_noise(min_freq=min_freq,
+                                        max_freq=max_freq,
+                                        n_samples=10000,
+                                        sampling_rate=sampling_rate,
+                                        amplitude=amplitude,
+                                        type=type)
+
+        noise.set_trace(long_noise, sampling_rate)
+        power_noise = self.tunnel_diode(noise)
+
+        power_mean = np.mean(power_noise)
+        power_std = np.std(power_noise)
+
+        return power_mean, power_std
+
     def has_triggered(self, channel):
         """
         Check if the detector system triggers on a given channel.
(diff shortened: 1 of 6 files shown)


NuRadioReco/modules/ARA/triggerSimulator.py
```diff
@@ -200,47 +247,44 @@ def run(self, evt, station, det,
...
         if not has_triggered:
             trigger.set_triggered(False)
             logger.info("Station has NOT passed trigger")
             trigger_time_sample = 0
-            station.get_trigger().set_trigger_time(trigger_time_sample / sampling_rate)
+            trigger.set_trigger_time(trigger_time_sample)
         else:
             trigger.set_triggered(True)
-            trigger.set_trigger_time(trigger_time_sample / sampling_rate)
+            trigger.set_trigger_time(trigger_time_sample)
```
cg-laser: you might want to rename the variable from `trigger_time_sample` to `trigger_time`. Otherwise its confusing.

daniel-zid:
> Changes look ok, but I have no way of testing it. But I guess you tested it so it's ok to merge. The old way anyway didn't work properly. Just the variable renaming would be good

I'm running some tests on the cluster, but it has been taken by the usual suspects and my jobs won't run. I might postpone merging until the tests are complete and we understand them.

NuRadioReco/modules/ARA/triggerSimulator.py
daniel-zid: Changed.

NuRadioReco/modules/ARA/triggerSimulator.py
```diff
@@ -200,49 +247,46 @@ def run(self, evt, station, det,
...
+        trigger_times = []
         for channel in station.iter_channels():
             channel_id = channel.get_id()
             if channel_id not in triggered_channels:
                 continue
             trigger[channel_id] = self.has_triggered(channel)
+            if trigger[channel_id]:
+                times = channel.get_times()
+                trace_after_diode = self.tunnel_diode(channel)
+                arg_trigger = np.argwhere( trace_after_diode == np.min(trace_after_diode) )[0,0]
```
christophwelling: Isn't this equivalent to just writing arg_trigger=np.argmin(trace_after_diode)?

NuRadioReco/modules/ARA/triggerSimulator.py
```diff
@@ -200,49 +247,46 @@ def run(self, evt, station, det,
...
-                tr = np.array(tr)
-                mask_trigger_in_coind_window = (tr >= i) & (tr < istop)
-                if(np.sum(mask_trigger_in_coind_window)):
-                    coinc += 1
-                    trigger_times.append(tr[mask_trigger_in_coind_window][0])  # save time/sample of first trigger in
-
-            if coinc >= number_concidences:
+        trigger_time = None
+
+        trace_times = station.get_channel(0).get_times()
```
christophwelling: There may not be a channel with ID 0.
As addressed in #160 , I don't think we should assume that all channels have the same trace_start_time, since there may be differences between the deep and shallow channels or if we have a double trigger for DnR

NuRadioReco/modules/ARA/triggerSimulator.py
daniel-zid: You're right. I've changed it.

NuRadioReco/modules/ARA/triggerSimulator.py
daniel-zid: That was in the module before. Let me change it.

daniel-zid: Keep in mind that the assumption that they have the same trace_start_time is nowhere. I get, for every channel, the correct trigger times array given by the get_times() method. I agree that the way of sliding the window is not the best, as it should go from the minimum time of all the channels, to the maximum time.

daniel-zid: Done. Let me know if it's better this way.

daniel-zid:
Here's a test of the effective volume: phased array vs ARA trigger. The phased array performs better at low energies, as expected, and if I had used equal bandwidths, I would have had a better improvement. However, the diode trigger has an increase at high energies for 100 m, and I don't understand why. Maybe it's a product of the non-linearity of the diode? No idea. I'm going to check if that depth has been calculated properly.
[ARA_vs_phased_Greenland.pdf](https://github.com/nu-radio/NuRadioReco/files/3846240/ARA_vs_phased_Greenland.pdf)

daniel-zid:
Ok, proof that the new ARA trigger works and it's worse than a phased array trigger. Simulations have been run with a large bandwidth. I'm going to run some simulations for 140-530 MHz, but I think the addressed changes and the checks are complete. @cg-laser @christophwelling do you approve the current version or do I need to change more things?

[ARA_vs_phased_Greenland.pdf](https://github.com/nu-radio/NuRadioReco/files/3857499/ARA_vs_phased_Greenland.pdf)

daniel-zid:
For a restricted band (140-530 MHz), the trigger is even better. Remember that this simulation has been made with the buggy arrival directions, but it is a fait trigger comparison nevertheless.

[band_ARA_vs_phased_Greenland.pdf](https://github.com/nu-radio/NuRadioReco/files/3863204/band_ARA_vs_phased_Greenland.pdf)

PR #167, opened 2019-11-25
Fix phased noise

daniel-zid:
It turns out that with the new proper noise normalisation, the noise-triggered events are more than what I'm comfortable with. I've modified the phased array examples to show how to create an acceptance window around the actual electric field, so that we don't consider the part of the traces which is only noise and we have fewer noise-triggered events as a consequence. I'm going to post some plots as soon as they're ready. I've also changed the bandwidth to the one we will actually use for Greenland and changed the order of one of the filters.

I thought about implementing this window inside the trigger module, but that's perhaps a bit too cumbersome. It would require the module to know either the noiseless traces and the noisy traces, or to restrict the traces to a user-defined window, which is more or less the same the new steering files do. I don't know what would be the best course of action here.


daniel-zid:
Here's a comparison of the phased array with the bandwidth given by Eric, a 1.5 sigma noiseless dipole and the ARA trigger. The dashed line represents the effective volume due to noise-triggered events, which is less than 1%.

[band_ARA_vs_phased_Greenland.pdf](https://github.com/nu-radio/NuRadioReco/files/3901869/band_ARA_vs_phased_Greenland.pdf)

PR #171, opened 2019-12-06
CoREAS reader fix

christophwelling:
Fixes 2 issues:
- readCoREASStation somehow didn't make it to Python3
- Convention was that readCoREAS an readCoREASStation always put the shower core at [0,0,0] and set the simStation position accordingly. Now that the simStation doesn't have its own position any more, the shower core is set to the right position relative to the station in the simShower.


christophwelling:
I had to add pos_northin and pos_easting to the example detector description. While doing that I also kicked out all the unused stations. There is no need to have such a complicated detector in the examples.

fschlueter:
Hi guys,
looks fine to me. I tried to use the same logic for the shower reader but realized that this does not work since the z-positions of the pulses in CoREAS are absolute. Therefore one would always get a core of (0, 0, 0) not taking into account the observation level. 
I spotted a potential bug, please check.
Cheers

NuRadioReco/detector/detector.py
```diff
@@ -403,6 +403,8 @@ def get_absolute_position(self, station_id):
             easting = res['pos_easting'] * unit_xy
         if(res['pos_northing'] is not None):
             northing = res['pos_northing'] * unit_xy
+        if(res['pos_altitude'] is not None):
+            altitude = res['pos_altitude'] * unit_xy
```
cg-laser: this wasn't a bug, please remove this change. The altitude is always saved in the default unit which is meters. The only exception is the special coordinate system at the South Pole where x/y is specified in feet.

cg-laser: And look at the docstring. The altitude is relative to the snow level at the time of the measurement. Which is a good thing for neutrino detectors,... but I can see that this is not ideal for cosmic rays. Not sure how to solve it... 
Maybe define two functions `get_absolute_position_wrt_surface` `get_absolute_position_absolute` hmm, but the naming is odd. `absolute_position` should be absolute. So we should rather rename the current function to `get_position_wrt_surface`

fschlueter: Ah sorry I did not catch that. Yes indeed we would need a function which returns the absolute position. Should I add this and rename the above function to `get_position_wrt_surface`?

fschlueter: Hi,
how do you think about this? two getter functions? I did not really understand why the function does not return any other altitude other than 0. You don't have snow accumulation? Just out of curiosity :)

anelles:
@cg-laser I think you can approve those changes now, right?

cg-laser:
no it's still wrong. The altitude is always in our default unit, so it does not need to be multiplied with `unit_xy`. 
And for the long run, we need to think about how to store altitude. Absolute or wrt to snow surface, ...

anelles:
@christophwelling can you stlll fix the changes requested by @cg-laser ?

christophwelling:
Not sure if I understood the issue correctly. @cg-laser Do we just remove the changes Felix made for now? In the long run we will need to define the positions/coordinate system properly, but I would prefer to not open that can of worms in this pull request.

fschlueter:
Hi of course I can  revert my changes but I think we need a proper solution already now. It seems that the code just work for the south pol right now.

christophwelling:
Yes, but this might be a bigger project if we want to do it right. I would prefer to merge this PR and solve the coordinate system issues in a different one.

fschlueter:
Okay, but how do i request the correct position for my case with the current code?

cg-laser:
Probably I wasn't clear enough. The only issue is that you interpreted the altitude as foot for a Southpole station, which is wrong. The altitude is always stored in meters. Because this is the default unit, we don't need to multiply with any unit. BTW: This is the logic, all quantities are in the database are  normally stored in the NuRadioReco default units. So there is no need for any conversion. 

I think with this fix, we can merge the PR and it should also work for you Felix right? Because we return the altitude (before we always returned 0). This is not a problem for existing code. The function is never used in NuRadioReco. It's only used in NuRadioMC where we anyway defined the altitude as 0 always. 

However, we now mean different things: For neutrino detectors we measure the altitude wrt to the ice surface, which is just 0 for the station center as the depth of the individual antennas are saved as relative distances to the station center. But for cosmic ray detectors it is the absolute height asl. 

I also noticed that the altitude of the ARIANNA stations is not set in the database. I'll correct that and update the JSON export.

fschlueter:
Ah indeed I didnt get this, sorry for delaying!

PR #172, opened 2019-12-11
normalize noise level to the correct sampling rate

cg-laser:
you might still need to fix the other examples. I didn't check


daniel-zid:
I have corrected all the examples in another pull request. Feel free to merge this one if you want.

cg-laser:
I added a more robust way of normalizing the noise level. What do you think?

PR #173, opened 2019-12-12
Envelope phased

daniel-zid:
New code for the envelope phased array. The principle is the same as the principle for the regular phased array, but removing the possibility of secondary channels, and applying the diode response to the phased array channels. A new diodeSimulator class has been created, which gives an easier access to the ARA diode model. A new module has been created in utilities as well, called traceWindows, which allows to specify a window to trigger on in order to reduce the number of noise-induced triggers. Examples have been created and updated, also for the regular phased array. Some improvements to the regular phased triggerSimulator have been made. A bug concerning a wrong noise amplitude due to a bad choice for the Nyquist frequency has been corrected.

I'm currently running some simulations to know how many noise-induced events there are for a typical simulation. If the number is negligible, I will drop the restrictive trigger window from the phased array examples.


NuRadioReco/modules/envelope_phasedarray/triggerSimulator.py
```diff
@@ -0,0 +1,220 @@
...
+        """
+        sampling_rate = station.get_channel(0).get_sampling_rate()
+        time_step = 1. / sampling_rate
+
+        diode = diodeSimulator(output_passband)
+
+        traces = {}
+        for channel in station.iter_channels():
+            channel_id = channel.get_id()
+            if channel_id not in triggered_channels:  # skip all channels that do not participate in the trigger decision
```
christophwelling: FYI: The iter_channels method has the use_channels parameter that does this automatically for you.

NuRadioReco/modules/phasedarray/triggerSimulator.py
```diff
@@ -122,28 +131,40 @@ def phased_trigger(self, station, beam_rolls, sec_beam_rolls, triggered_channels
...
+            trace = channel.get_trace()  # get the enveloped trace
+            times = channel.get_times()  # get the corresponding time bins
+
+            if cut_times != (None,None):
+                left_bin = np.argmin(np.abs(times-cut_times[0]))
+                right_bin = np.argmin(np.abs(times-cut_times[1]))
+                trace[0:left_bin] = 0
+                trace[right_bin:None] = 0
+
+            traces[channel_id] = trace[:]
```
christophwelling: Won't this change the traces?

cg-laser:
@daniel-zid can you update the modules to use the changes for noise normalization from #172 ? I just merged them.

daniel-zid:
The new way of normalising noise has been implemented. Git was complaining that one of the example files had conflicts, while no conflicts were shown. And after the merge, it did not let me push, so I had to force it. God knows.

daniel-zid:
It looks like Atom was laughing at my face and there were some conflicts after all. I'm going to fix it.

daniel-zid:
Alright, it looks like the problem is solved. I will merge the branch tomorrow if no one requests changes.

anelles:
Please check whether there are any interferences with the changes in #175 . Otherwise, please go ahead in merging.

daniel-zid:
As far as I know, we are going to have a trigger called envelope_trigger and another one called envelope_phased. Other than that, there would be no problems. Lilly could even use the passband attribute to specify the output bandpass filter of the diode. It seems that she has yet to write the code for the diode or the envelope she wants to use.

PR #176, opened 2020-01-10
An event reader for LOFAR

anelles:
Feature request from LOFAR.
Is clearly work in progress, but a pull request helps external collaborators to take a look.


anelles:
This will remain work in progress for now. Depending on when we find someone to work on this again.

PR #177, opened 2020-01-16
Adding copy statements so as not to modify the phased array trace

daniel-zid:
There was a mistake before when treating the cut times for the phased array trace. The trace was fetched from the channel by reference, and not copied, which means that the original channel trace was modified. This was not the case for the envelope phasing, though.

And I also added copy statements for the time arrays, just for the sake of being paranoid.


iplaisier:
Looks good, channel traces from phased array are not cut any more

PR #178, opened 2020-01-21
fix closing files (loop through items)

cg-laser:
bugfix in closing files


cg-laser:
@christophwelling ok like that?

PR #179, opened 2020-01-22
add new power integration trigger module

cg-laser:


cg-laser:
The test is failing because of the recent NuRadioMC change (pulse position in trace). @christophwelling can you update the reference? I'm not sure how to do it.

christophwelling:
I updated the NuRadioReco reference, but the NuRadioMC test is still failing. I don't know what's up with that, but it has to be fixed in NuRadioMC.
FYI: The comparison scripts in the tests have the option --create_reference. If you use that, instead of comparing the reco results to reference, the script will create a new reference.

anelles:
Actually in my pull request about the logging, the reference is fixed and
works ...

NuRadioReco/modules/trigger/powerIntegration.py
```diff
@@ -0,0 +1,136 @@
...
+    @register_run()
+    def run(self, evt, station, det,
+            threshold,
+            integration_window,
+            number_concidences=1,
+            triggered_channels=None,
+            coinc_window=200 * units.ns,
+            trigger_name='default_powerint'):
+        """
+        simulate simple trigger logic, no time window, just threshold in all channels
```
christophwelling: What do you mean by "no time window"?

NuRadioReco/modules/trigger/powerIntegration.py
```diff
@@ -0,0 +1,136 @@
...
+            trigger.set_triggered(False)
+            trigger.set_trigger_time(0)
+            logger.debug("station has NOT triggered")
+        station.set_trigger(trigger)
+
+        self.__t += time.time() - t
+
+    def end(self):
+        from datetime import timedelta
+        logger.setLevel(logging.INFO)
```
christophwelling: We should remember to adjust this to the new way the logger works. How we do that will depend on what gets merged first

NuRadioReco/modules/trigger/powerIntegration.py
cg-laser: The docu was wrong, I fixed it.

cg-laser:
It's again the Veff test. Restarting the test just worked.

PR #181, opened 2020-01-23
Change the general logging handling. 

anelles:
Issue being addressed:

We used the native logging in a strange way that never allowed to set globally whether one wanted debug output or not. Now all logger inherit from a NuRadioReco logger and can be set to debug in general.


anelles:
This touches a lot of files, so a small summary. Basically there is a function now in module.by setup_logger. This one should be called in any module sequence steering file with a globel log level setting. The rest is automatically. 
One can still set log levels of individual modules, however, this is not homogenized yet of where to do this.

anelles:
And what's with the tests again? I did not touch triggering stuff.

anelles:
I just checked: the master is also broken, the triggering test only. Is this collateral again from some NuRadioMC change? Because the test ran fine after the last merge.

anelles:
No discussion, great!

anelles:
I also edited a wiki page to explain the concept: https://github.com/nu-radio/NuRadioReco/wiki/Logging

christophwelling:
I added your wiki page to the documentation. I think it's nicer to have all information at one place.

PR #183, opened 2020-01-29
Eventbrowser overhaul

christophwelling:
Cleans up eventbrowser code, makes it run a bit faster by only loading the plots that are already shown.
It also adds a fancy new 3D plot.
This is all not done, but it also introduces some changes to NuRadioMC that I want to merge before tackling the multi-event issue to stop things from getting messy


cg-laser:
can you reset the eventid counter when a new file is opened? Currently, if the new file has fewer events it crashes.

anelles:
I think @cg-laser sort of reviewed in asking whether you can reset the counter. Otherwise, we cannot test, if it works.

christophwelling:
Sorry, I missed that :-)
The event_id is now reset when a new file is changed.
I also made it so you can only select 1 file at a time. Having multiple files caused errors if you closed a file.
Unfortunately, I also noticed a problem with the dataprovider: Since all plots try to access the dataprovider at the same time, deleting the NuRadioRecoio instance can cause an error if one function accessing the dataprovider deletes an instance that another function just created. I resolved this by not deleting the old instance, which works for now. It's a memory leak, but nothing severe unless you close and open too many files all the time.

cg-laser:
I actually used the new event browser to brows through ARIANNA data. It worked fine but I was probably not testing the new features that you implemented. 
I think to properly review this PR you'd need to provide an example file and/or show a few screenshots.

cg-laser:
I also trust Christoph that he did a good job. Nevertheless for those pull requests it would be nice to see a screenshot with the new features or an example file to have look by myself. Otherwise there is not much review going on except nothing breaks.

christophwelling:
Screenshot from 2020-02-05 09-50-21

Here is a screenshot of what it looks like. From a UI perspective the only big changes are the new 3D event display and the fact that the tabs have been changed to radio buttons. This way, only the elements that are currently shown are refreshed when changing to another event, which speeds thing up.
Most of the code changes are just putting the scripts for each plot into their own file to make the code readable again.
Unfortunately, GitHub won't let me upload .nur files here. Is there a good way to provide example files for pull requests?

cg-laser:
thanks for providing the screenshot. The plot looks great! 

I thought one can upload any file to the PR page... I have no good idea how to share the file then... maybe a private dropbox link? Or rename it to *.pdf, then you might be able to upload it ;-) just kidding but it actually might work

Feel free to merge the PR

PR #189, opened 2020-02-05
Vertex times

daniel-zid:
Trigger times now include the vertex times


daniel-zid:
Keep in mind that the trigger tests are going to fail because of the different times.

daniel-zid:
> we are already at a newer version. Can you update the changelog accordingly?
> 
> Why does the PA trigger sets the trigger time to the trace start time? Isn't there a specific location in the trace where the trigger is fulfilled?

No, there is not. When traces are phased, one must choose a reference position or a reference time, since the traces are moved around and the original timing is lost. In other words, the phased trace times do not correspond to any other channel times. I suppose we could choose a convention, if you want.

daniel-zid:
> we are already at a newer version. Can you update the changelog accordingly?
> 

The master still says 1.0.1, right?
> Why does the PA trigger sets the trigger time to the trace start time? Isn't there a specific location in the trace where the trigger is fulfilled?

cg-laser:
no, it's at 1.1 (https://github.com/nu-radio/NuRadioReco/blob/master/NuRadioReco/__init__.py). Sry this is my fault, I forgot to properly update the changelox.txt when I did the release.

daniel-zid:
The tests work perfectly on my machine but not on the build bot. I am lost. Could you take a look, @cg-laser?

anelles:
Have you tried recompling the ray tracer on your machine? Execute setup.py
again.
Anna

daniel-zid:
> Have you tried recompling the ray tracer on your machine? Execute setup.py again. Anna
> […](#)

Many many times

cg-laser:
seems that you solved it? It was just the wrong NuRadioReco version? I'd suggest merging NuRadioReco first, then changing travis back to master, and then merging this branch.

cg-laser:
ups I was confused. Hmm now we have this weird circular dependency between NuRadioMC and Reco...

cg-laser:
please continue this PR at #194

PR #193, opened 2020-02-07
add shower energy and CE profile to efield properties

cg-laser:
corresponds to https://github.com/nu-radio/NuRadioMC/pull/179

I think the tests fail because of the new efield parameters. Can we rewrite the test that only the parameters are being tested that are present in the reference file? Otherwise tests will always fail if we add a new parameter.


christophwelling:
Energy and CE profile are properties of the shower, so I think it should be saved in the shower class.

cg-laser:
I thought about that but is doesn't work because it is not guaranteed that each event has only one shower. The structure allows one event to have multiple showers, but there is not 1:n relation between a shower and the corresponding efields.

christophwelling:
Then we should implement a way to identify which shower an E-Field was derived from. We already have the shower energy in the shower and the station class. Now adding a third place where it might be just makes things even more confusing

cg-laser:
ok, I agree. This will make the event splitting more complicated though.

cg-laser:
@christophwelling can you review again? I removed the efield parameters. I only added saving of the config file. 
I also needed to update the reference because of the NuRadioMC change in the random numbers of the Askaryan modules.

cg-laser:
the test is failing because the branch doesn't use the correct NuRadioMC version. But this will be solved once the NuRadioMC changes get merged.

PR #196, opened 2020-02-18
Diode parameters

daniel-zid:
I noticed that the noise trigger rate for the envelope phased trigger was highly variable using the same configuration. The problem was that the calculation of the mean and standard deviation of thermal noise after passing through the diode needed more samples to be closer to the actual averages. With this pull request the noise trigger rate becomes more stable.

### DIFF
--- a/NuRadioReco/utilities/diodeSimulator.py
+++ b/NuRadioReco/utilities/diodeSimulator.py
@@ -126,7 +126,8 @@ class diodeSimulator:
                                    max_freq=1*units.GHz,
                                    amplitude=10*units.microvolt,
                                    type='rayleigh',
-                                   n_samples=10000):
+                                   n_tries=1,
+                                   n_samples=int(1e8)):
         """
         Calculates the mean and the standard deviation for the diode-filtered noise.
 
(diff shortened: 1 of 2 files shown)


NuRadioReco/utilities/diodeSimulator.py
```diff
@@ -150,20 +153,29 @@ def calculate_noise_parameters(self,
...
-        long_noise = channelGenericNoiseAdder().bandlimited_noise(min_freq=min_freq,
-                                        max_freq=max_freq,
-                                        n_samples=10000,
-                                        sampling_rate=sampling_rate,
-                                        amplitude=amplitude,
-                                        type=type)
+        power_mean_list = []
+        power_std_list = []
+
+        for i_sample in range(n_samples):
```
anelles: Is this just a handy coincidence that you do the calcualtion as many times as you have samples in the trace or am I missing the logic here?

anelles: Otherwise, I would suggest to add another parameter. The documentation is also a bit misleading now, as n_samples is both the loop and the number of samples, as the name suggests.

daniel-zid: They are different parameters. I'm going to change the name. Another option would be to increase the trace by a factor of 10000. I'm not sure which method would be faster, but in any case this has to be done only once and the loop seems just fine.

daniel-zid: The 10000 was there before for the ARA trace. I ignore if there was any rationale behind this number, but I chose to preserve it just in case.

daniel-zid: I've checked that using a single trace for this calculation makes my laptop suffer.

PR #199, opened 2020-02-20
Analog to digital converter

daniel-zid:
I've implemented a simple ADC model. It can be used to transform all the signals from a station to discretised values, but it also can be used for triggering only, without modifying the channel traces. This allows to use less power-hungry 8-bit amplifiers to decide if an event is stored and then digitising all the channels with a better 16-bit ADC (I don't remember what it's going to be used for RNO). The phased array and envelope phased array routines are already modified to do so. I've added new parameters for the detector files that are used. They are documented in the analogToDigitalConverter.py module. These are:

-"adc_reference_voltage", the maximum voltage for the ADC in volts
-The same fields for a normal ADC, but starting with "trigger_". These parameters are used for the triggering ADCs.

I've tested the code and it seems to work. I could show some examples and modify the standard tests to include discretisation in some of them.


NuRadioReco/detector/detector.py
```diff
@@ -94,7 +94,13 @@ def buffer_db(in_memory, filename=None):
                                 'adc_time_delay': channel['adcs.time_delay'],
                                 'adc_nbits': channel['adcs.nbits'],
                                 'adc_n_samples': channel['adcs.n_samples'],
-                                'adc_sampling_frequency': channel['adcs.sampling_frequency']})
```
cg-laser: this will likely break the database buffering as these fields are not present at the moment in the mysql database. I actually think that you don't need to change this function at all. We won't use mysql for RNO-G, and for ARIANNA we don't need this functionality. For a simulation study we anyway write the json file manually

daniel-zid:
> Can you easily add to also do the inverse? I.e. converting ADC counts back into voltages? The run method could get a flag to do so.
> 
The output of the ADC module are not ADC counts but a discretised voltage. It is not realistic but it can save lots of headaches during simulations. The counts can always be retrieved using the number of bits. A function can be added if you want either ADC counts or voltage, it's just that it seemed safer to me to use only voltages as a first step.

> Can you add a documentation of the adc database fields (and explain their meaning) to https://github.com/nu-radio/NuRadioReco/blob/master/NuRadioReco/autodoc/pages/detector_database_fields.rst?

I'll try tomorrow.

NuRadioReco/detector/detector.py
daniel-zid: I can remove that, no problem.

NuRadioReco/modules/analogToDigitalConverter.py
```diff
@@ -0,0 +1,318 @@
...
+        Parameters
+        ----------
+        evt: framework.event.Event object
+        station: framework.station.Station object
+        det: detector.detector.Detector object
+        random_clock_offset: bool
+            If True, a random clock offset between -1 and 1 clock cycles is added
+        adc_type: string
+            The type of ADC used. The following are available:
+            - perfect_floor_comparator
```
christophwelling: Doesn't it also allow perfect_ceil_comparator?

NuRadioReco/modules/analogToDigitalConverter.py
```diff
@@ -0,0 +1,318 @@
...
+                                                        output=output)
+
+            channel.set_trace(digital_trace, adc_sampling_frequency)
+
+        self.__t += time.time() - t
+
+
+    def end(self):
+        from datetime import timedelta
+        logger.setLevel(logging.INFO)
```
christophwelling: You should use self.logger here

NuRadioReco/modules/envelope_phasedarray/triggerSimulator.py
```diff
@@ -66,6 +71,9 @@ def envelope_trigger(self,
             the number of noise-induced triggers. Doing it the other way, that is,
             cutting and then filtering, will create two artificial pulses on the
             edges of the trace.
```
christophwelling: Please specify what cut_times means

NuRadioReco/modules/envelope_phasedarray/triggerSimulator.py
```diff
@@ -75,18 +83,35 @@ def envelope_trigger(self,
...
+                                        adc_type='perfect_floor_comparator',
+                                        diode=diode)
+                time_step = 1 / det.get_channel(station_id, channel_id)['trigger_adc_sampling_frequency']
+                times  = np.arange(len(trace), dtype=np.float)
+                times += channel.get_trace_start_time()
+
+            else:
+
+                trace = diode.tunnel_diode(channel)  # get the enveloped trace
+                times = np.copy(channel.get_times())  # get the corresponding time bins
```
christophwelling: times is already declared in line 96

NuRadioReco/modules/envelope_phasedarray/triggerSimulator.py
```diff
@@ -175,6 +207,9 @@ def run(self, evt, station, det,
             the number of noise-induced triggers. Doing it the other way, that is,
             cutting and then filtering, will create two artificial pulses on the
             edges of the trace.
```
christophwelling: Here as well: What does cut_times mean?

NuRadioReco/modules/phasedarray/triggerSimulator.py
```diff
@@ -131,6 +155,9 @@ def phased_trigger(self,
             Width of the time window used in the power integration
         cut_times: (float, float) tuple
             Times for cutting the trace. This helps reducing the number of noise-induced triggers.
+        trigger_adc: bool
```
christophwelling: Here too

NuRadioReco/modules/analogToDigitalConverter.py
daniel-zid: Changed

NuRadioReco/modules/analogToDigitalConverter.py
daniel-zid: Changed.

NuRadioReco/modules/envelope_phasedarray/triggerSimulator.py
daniel-zid: It is there on the previous file.

NuRadioReco/modules/envelope_phasedarray/triggerSimulator.py
daniel-zid: Changed.

daniel-zid:
Changes have been addressed.

daniel-zid:
Can we resurrect this thread, please? All the changes have been addressed and I need this branch to work on the new triggers.

PR #201, opened 2020-02-26
make check for duplicates optional

cg-laser:


cg-laser:
@christophwelling ok with you?

NuRadioReco/modules/triggerTimeAdjuster.py
```diff
@@ -50,6 +50,9 @@ def run(self, event, station, detector):
             trigger_time = trigger.get_trigger_time()
             for channel in station.iter_channels():
                 trigger_time_channel = trigger_time - channel.get_trace_start_time()
+                print(f"trace start time {channel.get_trace_start_time():.1f}")
```
anelles: Please change all the print statements to logging.

christophwelling:
It's parctically random which event will be returned by the get_event function and having event-dependent changes to the detector description may not work. Other than that, it should work, I think

cg-laser:
the NuRadioRecoio now issues a warning if an run/event id pair is requested that exists more than once.

PR #205, opened 2020-03-09
Fixes typo in stationParameters

christophwelling:


christophwelling:
Well, my code editor always suggested "inelasiticy" as first choice for a parameter name and that got incredibly annoying :-)

PR #207, opened 2020-03-10
Force polarization to be entirely eTheta or ePhi of eField reconstruction

ggaswint:
Added ability to force polarization to be entirely eTheta or ePhi in efield reconstruction using VoltageToEfieldConverter. Was added so that we can use voltageToEfieldConverter on all the dipoles only and get a reasonable reconstructed electric field. It was necessary because the cross-pol component of the dipoles is extremely small, which caused the ePhi component to be artificially enlarged by many orders of magnitude larger than the eTheta component.


NuRadioReco/modules/voltageToEfieldConverter.py
```diff
@@ -136,6 +136,10 @@ def run(self, evt, station, det, debug=False, debug_plotpath=None, use_channels=
             be save into the `debug_plotpath` directory
         use_channels: array of ints
             the channel ids to use for the electric field reconstruction
+        use_MC_direction:
```
anelles: I know you didn't add the parameter, but can you document it as well, please.

anelles:
I assume the physics is well tested. Code looks good, but I cannot test it of course.

NuRadioReco/modules/voltageToEfieldConverter.py
ggaswint: I have added this documentation.

PR #211, opened 2020-03-25
Bug fixes

ggaswint:
1) Removed function "get_relative_positions(self, station_id)" from detector.py because the ordering of channels is different from a sql initiated detector compared to a json initiated detector. Must use "get_relative_position(self, station_id, channel_id)", which specifies which channel to get the position for.

2) Fixed correlationDirectionFitter and beamFormingDirectionFitter to use "get_relative_position" instead of "get_relative_positions", see previous comment.

3) Added "self."" to all occurrences of "logger" in channelTimeWindow.py


ggaswint:
Also added some documentation to channelTimeWindow.py

NuRadioReco/modules/correlationDirectionFitter.py
```diff
@@ -23,12 +24,11 @@ def __init__(self):
         self.__azimuth = []
         self.__delta_zenith = []
         self.__delta_azimuth = []
-        self.logger = logging.getLogger('NuRadioReco.correlationDirectionFitter')
         self.begin()
 
     def begin(self, debug=False, log_level=None):
         if(log_level is not None):
-            self.logger.setLevel(log_level)
+            logger.setLevel(log_level)
```
cg-laser: can you change the logger back to `self.logger`. Also don't call `basicConfig()`.

NuRadioReco/modules/correlationDirectionFitter.py
```diff
@@ -218,7 +216,7 @@ def ll_regular_station_fft(angles, corr_02_fft, corr_13_fft, sampling_rate, posi
             t02s = toffset[indices]
             ax.plot(toffset[indices], corr_02[indices], 'o')
             imax = np.argmax(corr_02[indices])
-            self.logger.debug("offset 02= {:.3f}".format(toffset[indices[imax]] -  (delta_t_02 / sampling_rate)))
```
cg-laser: change back to use logger

changelog.txt
```diff
@@ -38,5 +39,8 @@ bugfixes:
 -Fixes increase in filesize caused by switch to python3
 -Fixed bug when using no secondary channels on the phased array
 -Fixed bug when using ARA trigger
+-Removed function "get_relative_positions(self, station_id)" from detector.py because the ordering of channels is different from a sql initiated detector compared to a json initiated detector. Must use "get_relative_position(self, station_id, channel_id)", which specifies which channel to get the position for.
+-Fixed correlationDirectionFitter and beamFormingDirectionFitter to use "get_relative_position" instead of "get_relative_positions", see previous comment.
```
cg-laser: I think the second and third point is too detailed. It is clear that modules might need adjustment. 
Also, you should add the changes to the top of this file under "version 1.1.2 -"

changelog.txt
```diff
@@ -30,6 +31,7 @@ new features:
 -Trigger times now include the time with respect to the first interaction (vertex times)
 -Analog to digital converter module added
 -Improved calculation of the diode noise parameters
+-Added the ability to force polarization to be only eTheta or ePhi in the reconstruction of the electric field using voltageToEfieldConverter.py
```
cg-laser: I think this also needs to go up to the most recent version.

ggaswint: Yeah I wasn't sure if the dates represent start times or end times for that particular version. But I had moved it to version 1.1.2

PR #214, opened 2020-04-06
Convert datetime

christophwelling:
The way datetime was converted to astropy time breaks with astropy version 4.0. This one works for both 4.0 and 3.0.


christophwelling:
Astropy 4.0 is what made this change necessary. Astropy 3.0 works fine.
I agree that it is ugly, but I don't know a more elegant way to fix this. The best solution would be to strongly discourage using datetime, but even then we still need this to read old files

PR #219, opened 2020-04-22
Alias

daniel-zid:
Simulations with alias (higher Nyquist zones) are now supported. A couple of bugs have been taken care of, such as:

- ADC resampling is now correctly implemented
- Noise trigger rate modules now use the correct window widths


NuRadioReco/modules/analogToDigitalConverter.py
```diff
@@ -254,18 +282,63 @@ def get_digital_trace(self, station, det, channel,
...
+            filtered_trace = butterworth_filter_trace(trace, input_sampling_frequency,
+                                                      passband)
+        else:
+            filtered_trace = trace[:]
+
+        # Random clock offset
         delayed_times = times + adc_time_delay
-        interpolate_trace = interp1d(times, trace, kind='quadratic',
-                                     fill_value='extrapolate')
+        interpolate_trace = interp1d(times, filtered_trace, kind='quadratic',
```
cg-laser, line 307: did you check that quadratic interpolation does not create artifacts, e.g., a quick significant over/under shooting of the values? Wouldn't a linear interpolation be better? 
An alternative (and likely saver approach) would be to upsample such that the downsampling factor is an integer. Then you can just take a subselection of the data points. E.g. if the downsampling factor is 2, just take very other sample.

NuRadioReco/modules/io/eventWriter.py
```diff
@@ -17,7 +17,7 @@ def get_header(evt):
         if(station.has_sim_station()):
             header['stations'][station.get_id()]['sim_station'] = {}
             header['stations'][station.get_id()]['sim_station'] = station.get_sim_station().get_parameters()
-    header['stations'][station.get_id()][stnp.station_time] = station.get_station_time()
+        header['stations'][station.get_id()][stnp.station_time] = station.get_station_time()
```
cg-laser, line 20: thanks for spotting this. It seems that it never worked for multiple stations

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -111,3 +112,116 @@ def get_electric_field_energy_fluence(electric_field_trace, times, signal_window
...
+    Returns
+    -------
+    filtered_spectrum: array of complex
+        The filtered spectrum
+    """
+
+    f = np.zeros_like(frequencies, dtype=np.complex)
+    mask = frequencies > 0
+    b, a = butter(order, passband, 'bandpass', analog=True)
+    w, h = freqs(b, a, frequencies[mask])
```
cg-laser, line 222: can you change the imports to `from scipy import signal.freqs` such that you need to call this function via `signal.freqs` or even better via `scipy.signal.freqs`. `freqs` is common variable name that will just override the function. So leaving it like this can be error prone.

changelog.txt
```diff
@@ -7,6 +7,7 @@ version 1.1.2 -
 new features:
 - Added the ability to force polarization to be only eTheta or ePhi in the reconstruction of the electric field using voltageToEfieldConverter.py
 - new detector which allows to change certain parameters of the detector description to study systematic uncertainties.
+- Simulations with alias (higher Nyquist zones) now supported
```
cg-laser, line 10: specify that this only applies to the phased array trigger simulator (I guess)

daniel-zid:
> I had a brief look at the code. As far as I can say it looks good except for the inline comments I made.
> I didn't check the implementation of the phased array module thoroughly. I suggest that you ask someone who knows the details of this trigger to review the implementation (e.g. Cosmin).

Hi Christian. Thank you for the quick review. I will implement the requested changes as soon as I can. I'm going to ask Cosmin to review it as well.

NuRadioReco/utilities/trace_utilities.py
daniel-zid: Changed.

changelog.txt
daniel-zid: Changed.

NuRadioReco/modules/analogToDigitalConverter.py
daniel-zid: Also, I've forgotten to comment something important. I cannot use a standard Fourier resampling method because it removes every component above half the Nyquist frequency, effectively destroying all the information in the higher Nyquist zones.

cg-laser: yes, downsampling wouldn't work, but upsampling should be possible, shouldn't it? 
I guess my proposed method would work, but only for discrete time shift. E.g. if you upsample by a factor 10, you could choose the start sample from 0-9 and then take every 10th time bin. 
Maybe/probably the interplation method works better though. If you checked that the quadratic interpolation does not do weird stuff it might be the best. I just saw sometimes a weird behavior with higher order interpolations.

cozzyd: I think upsampling followed by interpolation is the way to go. Interpolation causes phase issues at high frequencies and can add power at new frequencies for higher-order polynomials.  In my experience, upsampling by a factor of 4 and then linearly interpolating works decently well although you can of course do some numerical tests. 

If you're just trying to shift every sampled time by the same non-integer amount though, you can also use the Fourier Shift theorem (you just have to be careful around the boundaries since it's cyclic).   Basically a delay of \tau is in the time domain domain corresponds to multiplying by  exp(-j \omega \tau) in the Fourier domain (this is obvious if you think about the definition of group delay). But you'd want to either truncate the signal or add some zeroes (and then remove them) to avoid wraparound problems.  

Actually, unless you're losing a sample, you're going to have a boundary problem here too right?

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -111,3 +112,53 @@ def get_electric_field_energy_fluence(electric_field_trace, times, signal_window
...
+        logger.warning(warning_msg)
+
+    int_factor = int(int_factor)
+
+    if (int_factor <= 1):
+        error_msg = "Upsampling factor is less or equal to 1. Upsampling will not be performed."
+        raise ValueError(error_msg)
+
+    n_zeroes = int_factor - 1
+    zeroed_trace = []
```
cozzyd, line 150: This is fine for now but if performance matters later, it would be much better to make a zeroed trace first and fill in the non-zero samples (so you don't have so many needless memory allocations).

NuRadioReco/modules/analogToDigitalConverter.py
daniel-zid: > I think upsampling followed by interpolation is the way to go. Interpolation causes phase issues at high frequencies and can add power at new frequencies for higher-order polynomials. In my experience, upsampling by a factor of 4 and then linearly interpolating works decently well although you can of course do some numerical tests.
> 
> If you're just trying to shift every sampled time by the same non-integer amount though, you can also use the Fourier Shift theorem (you just have to be careful around the boundaries since it's cyclic). Basically a delay of \tau is in the time domain domain corresponds to multiplying by exp(-j \omega \tau) in the Fourier domain (this is obvious if you think about the definition of group delay). But you'd want to either truncate the signal or add some zeroes (and then remove them) to avoid wraparound problems.
> 

Yes, I think that would be better. I'll give it a try.
> Actually, unless you're losing a sample, you're going to have a boundary problem here too right?

That's right. But it's no big deal losing a sample. Or two or three. Voltages are long enough most of the time.

NuRadioReco/utilities/trace_utilities.py
daniel-zid: You're right. I can change that.

NuRadioReco/utilities/trace_utilities.py
```diff
@@ -111,3 +112,53 @@ def get_electric_field_energy_fluence(electric_field_trace, times, signal_window
...
+    for point in trace[:-1]:
+        zeroed_trace.append(point)
+        for zero in range(n_zeroes):
+            zeroed_trace.append(0.)
+
+    upsampled_delta_time = 1 / (int_factor * original_sampling_frequency)
+    upsampled_times = np.arange(0, len(zeroed_trace) * upsampled_delta_time, upsampled_delta_time)
+
+    cutoff = 1./int_factor
+    fir_coeffs = firwin(ntaps, cutoff)
```
cozzyd: Does the window choice matter here? Probably not very much but I think the way we've traditionally upsampled (zero-pad in frequency domain) is more akin to a rectangular window if I understand how firwin works.

daniel-zid: I have no experience working with FIRs, so I've just used the default values. I can change it to `window=boxcar` if you like.

daniel-zid: But yeah, I think that if we want the actual taps we need to use 'boxcar'. I didn't even realise. Let me try.

daniel-zid: Well, waveforms look ok with the boxcar window. Might as well keep it.
Figure_1

NuRadioReco/examples/AliasPhasedArray/Noise_trigger_rate/noise_trigger_rate.py
```diff
@@ -0,0 +1,303 @@
...
+    elif adc_type == 'perfect_ceiling_comparator':
+        digital_trace = perfect_ceiling_comparator(resampled_trace, adc_n_bits,
+                                                   adc_ref_voltage, output)
+
+    # Upsampling with an FIR filter (if desired)
+    if upsampling_factor is not None:
+        if (upsampling_factor >= 2):
+            upsampling_factor = int(upsampling_factor)
+            upsampled_trace = upsampling_fir(digital_trace, adc_sampling_frequency,
+                                             int_factor=upsampling_factor, ntaps=2**7)
```
cozzyd: ntaps probably needs to be a function of the upsampling_factor? (or configurable)

cozzyd: I guess I commented on the wrong commit, but same principle stands if you want this to be a little more general :).  Basically the reason is that the number of taps needed will scale with the upsampling factor because the taps are calculated on the upsampled waveform.

NuRadioReco/utilities/trace_utilities.py
cozzyd: yeah, I I'm not sure what the best choice is!  I guess if the window is big enough it might not matter.

cozzyd:
Quick readthrough looks sensible. I commented on a few things but they're probably not so important.

NuRadioReco/examples/AliasPhasedArray/Noise_trigger_rate/noise_trigger_rate.py
daniel-zid: I've used 16 because that's what you told me. Is 16 fine for an upsampling of 4 or less? After that, how should I scale?

cozzyd: If you've done tests and showed that it works well for that window length then that's probably fine!  However, it probably needs to scale with the upsampling factor (unless firwin automatically scales it or something).  Maybe it's a moot point if higher upsampling factors are never used?  If the window length is ever shorter than the number of zeroes stuffed between samples, though, that will cause weird problems.  Maybe upsampling_factor * 4 isn't a terrible rule of thumb?  You can also just sidestep it by making it an argument, since it could be an important parameter to vary in a study.

NuRadioReco/modules/analogToDigitalConverter.py
daniel-zid: I don't think upsample is necessary. The input signals are already at 5 GHz, so I think that's more than enough for an interpolation. It would be nice to make sure that every input signal has a sampling rate of 5 GHz or greater, and if not, upsample the trace.

cozzyd: agreed, as long as the sample rate is a few times higher than the highest frequency of interest then there's no need to upsample.

NuRadioReco/utilities/trace_utilities.py
daniel-zid: Changed.

NuRadioReco/examples/AliasPhasedArray/Noise_trigger_rate/noise_trigger_rate.py
daniel-zid: I have implemented your rule of thumb. The number of taps can be specified now using the detector configuration, in the same way as ADC sampling frequency or reference voltage.

NuRadioReco/modules/analogToDigitalConverter.py
daniel-zid: Now the delay is carried out using phases in Fourier space and the resulting trace is cropped. I've adjusted a bit the timing in the ADC module but ultimately it's not that relevant because it's just a global offset.

daniel-zid:
Alright, I've implemented the phased delay, now the number of taps can be included in the detector configuration file, and I've also changed the way to fill the zeroed trace. Let me know if you have any comments.

daniel-zid:
I found out that there was a bug in the noise trigger rate calculation. I'm rerunning some simulations to get the correct trigger rate. In principle this should not change much the Nyquist zone comparison, as the thresholds I was using were for tens of Hz instead of a few Hz as I expected, which means the efficiency of the three Nyquist zones were overestimated. And the second and third zones were underperforming anyway.

daniel-zid:
@cg-laser @cozzyd All the requested changes have been addressed. Could someone approve the pull request or tell me if there's anything left to change? Thanks!

cozzyd:
It could be that I don't properly know how to use the PR feature on Github, but I think @cg-laser has to approve that his requested changes were made?

PR #220, opened 2020-04-28
hl-trigger coincidence window

cg-laser:
changes hl trigger to count coincidence window from point of threshold crossing only and not from all points above threshold

This seems to be closer to how the hardware works.


anelles:
Pull looks good, but it probably needs an update from the master to make the tests pass again.

anelles (approved):
Sorry that became dormant. Looks still good.

anelles:
@cg-laser, merge?

cg-laser:
this will break all tests... I think it would be better to make a new hl trigger module with this change.

PR #221, opened 2020-05-07
Require specific tinydb version for tests

christophwelling:
Sets tinydb version to 3.15.2


cg-laser:
can you also adapt the pyproject.toml file?

christophwelling:
Done. While I was on it I also updated the required python version

PR #226, opened 2020-06-02
setting antenna model by depth is now optional and can be deactivated

cg-laser:
So far the depth of the antenna was used to automatically determine if the antenna is in firn or in air. This PR makes this option optional which is useful for simulation studies. It will be made the default in NuRadioMC


daniel-zid:
It looks fine and it works on my machine.

If we're going to call the detector with antenna_by_depth=False from NuRadioMC, shouldn't we adapt the generic detector as well?

daniel-zid:
It looks good!

PR #230, opened 2020-06-04
Fail on nonexistent antenna model

cg-laser:
It is quite dangerous that NuRadioReco uses an antenna pattern of unity if no antenna pattern was found. Now NuRadioReco raises an error if the antenna pattern was not found.


daniel-zid:
It looks alright to me. Only the tests should be changed.

daniel-zid:
I think that someone is me. I'm sorry.

PR #237, opened 2020-07-07
RNO detector

christophwelling:
Adds a detector description for an RNO station and updates the RNO amplifier response to the most recent measurements.


NuRadioReco/detector/RNO_G/full_RNO_station.json
```diff
@@ -0,0 +1,356 @@
...
+            "station_id": 101
+        }
+    },
+    "positions": {},
+    "stations": {
+        "01": {
+            "pos_altitude": 0.0,
+            "pos_easting": 0.0,
+            "pos_northing": 0,
+            "pos_site": "greenland",
```
lpyras, line 351: When do we use Greenland and when Summit?

PR #240, opened 2020-07-10
Replace tinydb.purge calls

christophwelling:
The latest tinydb version renamed the purge() method to truncate(). When I changed that, I apparently missed some spots.


anelles (approved):
Thanks for fixing. Looks good. Do we need to move up the required version for tinydb or does this also work with older versions?

christophwelling:
No, the required version was changed a while back, I just missed a spot.

PR #242, opened 2020-07-20
Detector browser

christophwelling:
Adds a visualization for detector descriptions, similar to the event browser.


anelles (approved):
If you could add an image of the detector view for documentation purposes (I have seen it of course from you) then I think we are ready to merge. 
Even though in all the plotting things, the functions do not have doc strings ... but I guess this has been the standard anyway.

PR #243, opened 2020-07-23
Save git hashes

christophwelling:
Using GitPython (https://gitpython.readthedocs.io/en/stable/) we can get the git hash of NuRadioReco, NuRadioMC and radiotools and store them as an event parameter. This way, we can later tell which exact software version a .nur file was created with

### DIFF
--- a/NuRadioReco/framework/event.py
+++ b/NuRadioReco/framework/event.py
@@ -4,6 +4,7 @@ import NuRadioReco.framework.station
 import NuRadioReco.framework.radio_shower
 import NuRadioReco.framework.hybrid_information
 import NuRadioReco.framework.parameters as parameters
+import NuRadioReco.utilities.version
 from six import itervalues
 import collections
 import logging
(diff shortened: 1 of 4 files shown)


cg-laser:
We already have this feature for NuRadioMC with the additional feature that also the version number is saved. It's implemented in the utilites: https://github.com/nu-radio/NuRadioMC/blob/master/NuRadioMC/utilities/version.py and saved in the hdf5 output files. I think it is a good idea to also save it to the nur files. I suggest to move the utility function over to NuRadioReco and use these functions. As we already have a implementation that does not need an external package, I suggest we keep using it to reduce the number of dependencies.

christophwelling:
Nice, I didn't know about that. I moved the utility over to NuRadioReco and will change that in NuRadioMC once this PR is merged, so I don't break anything over there. The implementation seems to have broken when we switched to py3 anyway.

anelles (approved):
Looks good. Please don't forget to update this in NuRadioMC like you said.

PR #245, opened 2020-07-27
allow for a different bandpass filter per channel

cg-laser:
implement that the user can specify a different bandwidth per channel


cg-laser:
I'm not sure how useful this option is. One can always just specify a bandwidth larger than the nquist frequency so apply no filter. 
Also for NuRadioMC the function always needs to return a filter function so that the bandwidth can be determined properly.

PR #246, opened 2020-07-31
Dynamic signal window

christophwelling:
Having a static signal and noise window for all events isn't practical, since our traces are so short that the signal can be practically anywhere in the trace.
This PR lets the channelSignalReconstructor put the signal window around the trace maximum and set the noise window as anything outside the signal window.
The parameters used to specify signal and noiose windows are changed from start and end times to start times and window length.


christophwelling:
This could cause problems if the pulse is right at the start of the trace. What would be the use case for this feature?

cg-laser:
my experience is that the end of the trace is often more noisy then the beginning. Therefore it is saver to estimate the noise RMS from the part of the trace before the signal. But I agree that this can be tricky depending on where the signal is.

anelles:
I think we should merge this for now and then revisit this when we have concrete evidence that it is not working well as it is. Since we are mostly dealing with simulations right now (or traces that are almost too short for a noise window (i.e. ARIANNA)), I don't think it is a show-stopper at this point.

anelles (approved):

PR #247, opened 2020-08-02
change random generator in readCoREAS to avoid impacting the global n…

lpyras:
I changed the random generator in the readCoREAS file from np.random.seed(seed) to np.random.RandomState(seed). The default in is now seed=None. RadomState has the advantage, that it avoids impacting the global numpy state.


NuRadioReco/test/tiny_reconstruction/reference.json
```diff
@@ -1 +1 @@
...
..._amplitude_envelope": [0.2803517851099761, 0.2802588963924389, 0.26371616474120246, 0.28776646182968296], "P2P_amplitude": [0.5013210481362056, 0.5560835859923348, 0.5205909728639548, 0.551617830513689], "cr_xcorrelations": [null, null, null, null], "nu_xcorrelations": [null, null, null, null], "signal_time": [636.2440214285433, 634.0440214285433, 623.2440214285433, 622.8440214285433]}, "electric_field_parameters": {"ray_path_type": [null, null], "polarization_angle": [1.467057253263737, 1.4890983521441759], "polarization_angle_expectation": [1.324985858091711, -1.816606795498082], "signal_energy_fluence": [[0.0, 3.6371878455679414, 335.5500393638057], [0.0, 2.8768866244517146, 429.10608047739856]], "cr_spectrum_slope": [null, -1.7427483474721464], "zenith": [0.7862752532958984, 0.7862752532958984], "azimuth": [3.9598489379882813, 3.9598489379882813], "signal_time": [574.2189260067184, null], "nu_vertex_distance": [null, null], "nu_viewing_angle": [null, null], "max_amp_antenna": [null, null], "max_amp_antenna_envelope": [null, null], "reflection_coefficient_theta": [null, null], "reflection_coefficient_phi": [null, null], "cr_spectrum_quadratic_term": [null, 0.5690068495080591]}}}
```
anelles, line 1: It is super hard to tell in this json format. @lpyras did you check that the changes were small or explainable by the changes in a consistent way? Please add a general comment so I can get to approving this PR.

anelles:
I find it very curious that these changes change only the Single Events at South Pole. Something seems to be off about this. I am happy to approve this PR, but when changing random state settings, we have to be careful to not mess up more general properties. @christophwelling could you take a deeper look?

christophwelling:
This is because only the SP test includes noise. The random seed in that test is set by np.random.seed(0), but now that the channelGenericNoiseAdder has its own randomState, it is not affected by that any more and no seed is set. So once we set the seed correctly in the NuRadioMC test, it should fix it. I'm going to take a look at that.
We had the same issue with some of the NuRadioReco test (plus interplay between corREASReader and noiseAdder). Lilly tested it very carefully, so I'm pretty sure the change does what it is supposed to do.

christophwelling:
I changed the test to uses a NuRadioMC branch where the random seed is set according to these changes. Now the test works again

lpyras:
> I changed the test to uses a NuRadioMC branch where the random seed is set according to these changes. Now the test works again

Thanks for doing this!

christophwelling:
You're welcome. I'll leave you the honor of merging the PR :-)

PR #249, opened 2020-08-07
Removes obsolete parameters in simStation init

christophwelling:
Removes some parameters in the simStation __init__ method that were left over from the time when stations still contained traces

### DIFF
--- a/NuRadioReco/framework/sim_station.py
+++ b/NuRadioReco/framework/sim_station.py
@@ -12,12 +12,10 @@ logger = logging.getLogger('SimStation')
 
 class SimStation(NuRadioReco.framework.base_station.BaseStation):
 
-    def __init__(self, station_id, sampling_rate=None, trace=None):
+    def __init__(self, station_id):
         NuRadioReco.framework.base_station.BaseStation.__init__(self, station_id)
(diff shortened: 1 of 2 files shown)


NuRadioReco/framework/sim_station.py
```diff
@@ -12,12 +12,10 @@
 
 class SimStation(NuRadioReco.framework.base_station.BaseStation):
 
-    def __init__(self, station_id, sampling_rate=None, trace=None):
+    def __init__(self, station_id):
         NuRadioReco.framework.base_station.BaseStation.__init__(self, station_id)
         self.__magnetic_field_vector = None
         self.__simulation_weight = None
-        if(trace is not None and sampling_rate is not None):
```
anelles, line 19: This is removed because we now don't store the electric field directly in the simStation anymore, right? Just checking.

anelles (approved):
If you can answer my one question @christophwelling, then I am happy for this to be merged.

NuRadioReco/framework/sim_station.py
christophwelling: Yes, we must have missed this part when we introduced the dedicated E-Field class

PR #250, opened 2020-08-19
Adds automatic linting for PRs

christophwelling:
This should automatically run a linter for every new pull request


anelles:
I have to admit that I had to google what this does ... but nice initiative. I like it. Not sure we always want to lint the whole code base though. How about we do this once and then only what is new in the PR?

christophwelling:
Yes, I will set it back to only lint changed files once this is done. But I first want to make sure that all the code is in a good state. Same goes for linting on every commit: This is useful now for making sure everything works, but I can set it back once I'm done.

NuRadioReco/detector/antennapattern.py
```diff
@@ -103,7 +97,7 @@ def get_group_delay(vector_effective_length, df):
     return -np.diff(np.unwrap(np.angle(vector_effective_length))) / df / units.ns / 2 / np.pi
 
 
-def parse_WIPLD_file(ad1, ra1, orientation, gen_num=1, s_paramateres=[1, 1]):
+def parse_WIPLD_file(ad1, ra1, orientation, gen_num=1, s_parameters=None):
```
cg-laser, line 100: why is this being changed by the linter

NuRadioReco/detector/antennapattern.py
```diff
@@ -67,9 +61,9 @@ def interpolate_linear_vectorized(x, x0, x1, y0, y1, interpolation_method='compl
     mask = x0 != x1
     result = np.zeros_like(x, dtype=np.complex)
     denominator = x1 - x0
-    if(interpolation_method == 'complex'):
+    if interpolation_method == 'complex':
```
cg-laser, line 64: why are the parentheses removed?

NuRadioReco/detector/filterresponse.py
```diff
@@ -95,44 +92,44 @@ def get_filter_response(frequencies, filter_name):
 if __name__ == "__main__":
     from matplotlib import pyplot as plt
 
-    ff = np.linspace(10 * units.MHz, 500 * units.MHz, 100)
-    df = ff[1] - ff[0]
+    freqs = np.linspace(10 * units.MHz, 500 * units.MHz, 100)
```
cg-laser, line 95: why is this being changed?

NuRadioReco/detector/antennapattern.py
christophwelling: I had to look this complaint up, too: If the default parameter of a function is something like a list or a dict, the list will be created the first time the function is called. Any function calls after that don't create a new list, but reference the first one. So if anything causes that list to change, the function will be called with a different default value. Therefore, Linters complain about something like this.
In this case, it's probably not too bad, since the parameters should not change, but better safe than sorry.

NuRadioReco/detector/antennapattern.py
christophwelling: I thought this was part of pep8, but I checked and it doesn't seem to be explicitly stated. If you prefer, we can get rid of this check.

NuRadioReco/detector/antennapattern.py
cg-laser: yes that makes sense.

NuRadioReco/detector/filterresponse.py
christophwelling: ff is also used as a parameter name for some of the functions that are defined in the same file.

NuRadioReco/detector/antennapattern.py
cg-laser: yes I'd remove it. I personally find it more readable with parentheses.

cg-laser:
@christophwelling sorry for my ignorance, but I still don't see where I can see the linter results. Can you add a short description and/or post a screenshot?

christophwelling:
linter_screenshot

cg-laser:
thanks, and we don't see any errors because you already fixed everything?

christophwelling:
Yes. The older commits have plenty of errors ;-)

cg-laser:
Got it thanks,
is there a way to only show the errors?

christophwelling:
As far as I know not, but it is set to only check files that were changed in the pull request. So this should not be an issue for other PRs

cg-laser (approved):
I did not go through all the changes but what I went through looks good. I also checked on a few files that my autoformatter agrees with the changes ;-) Thanks a lot for doing all this cleanup work. 

I saw that you also fixed several typos. Does the linter find them automatically?

christophwelling:
The linter doesn't, but PyCharm has a spell check. Most of its complaints are not very useful, but sometimes it actually finds something.

cg-laser:
@christophwelling shall we merge this one?

christophwelling:
I would like to merge the fourier shift PR first and make sure everything works. Can you approve that one?

PR #253, opened 2020-08-31
Fix bug voltage to efield converter

ggaswint:
method get_array_of_channels() in voltageToEfieldConverter.py had an if statement that required a sim_station to run. I first added an if statement to check if a sim_station exists, so that this code could be executed without requiring a sim_station


christophwelling:
Thanks for catching the problem. Unfortunately, with these changes, it will always use the index of refraction of air if no simStation exists, which is not necessarily what we want. I would suggest to instead check for is_cosmic_ray on the station instead of the simStation. This would require the user to always set the event type in the station, but that is something you are supposed to do anyway.

ggaswint:
@christophwelling I have made the changes as you requested. The if check now looks for a particle type from the base station.

christophwelling (approved):

PR #254, opened 2020-09-01
Fix backward compatibility

christophwelling:
Ensures packward compatibility of the internal looping changes


cg-laser (approved):
thanks for fixing

PR #256, opened 2020-09-07
Hides astropy Erfa warnings

christophwelling:
Disables all astropy warnings of the class ErfaWarning. This gets rid of the 'dubious year' warning.


cg-laser (approved):
awesome. Thanks!

PR #259, opened 2020-09-11
Refactor phased array

cg-laser:
Refactors the phased array trigger module to provide a function that can be called on traces to calculate the trigger. This is required to use this code to calculate the thermal noise trigger rate. 

This PR also adds a chebychev filter to the bandstopfilter module


NuRadioReco/modules/phasedarray/triggerSimulator.py
```diff
@@ -18,265 +19,284 @@
 default_sec_angles = np.insert(default_sec_angles, len(default_sec_angles), default_angles[-1] + 3.5 * units.deg)
 
 
-class triggerSimulator:
+def get_antenna_positions(station, det, triggered_channels=None, component=2):
```
clark2668, line 22: Needs docs (purpose, arguments, returns, etc.), and I think "component" as the argument is maybe not so clear? Also, I'm not sure why a function is needed just to return the depths of the antennas, so an explanation added would be good?

NuRadioReco/modules/phasedarray/triggerSimulator.py
```diff
@@ -18,265 +19,284 @@
...
+            roll = int(np.round(delay / time_step))
+            subbeam_rolls[channel_id] = roll
+        logger.debug("angle:", angle / units.deg)
+        logger.debug(subbeam_rolls)
+        beam_rolls.append(subbeam_rolls)
+
+    return beam_rolls
+
+
+def get_channel_trace_start_time(station, triggered_channels):
```
clark2668, line 85: Needs docs. Also, I'm not sure I see the point of the get_channel_trace_start_time? An explanation about why this careful checking is needed would be good.

NuRadioReco/modules/phasedarray/triggerSimulator.py
```diff
@@ -18,265 +19,284 @@
...
+            channel_trace_start_time = channel.get_trace_start_time()
+        elif channel_trace_start_time != channel.get_trace_start_time():
+            error_msg = 'Phased array channels do not have matching trace start times. '
+            error_msg += 'This module is not prepared for this case.'
+            raise ValueError(error_msg)
+
+    return channel_trace_start_time
+
+
+def check_vertical_string(station, det, triggered_channels):
```
clark2668, line 100: Needs docs.

NuRadioReco/modules/phasedarray/triggerSimulator.py
```diff
@@ -18,265 +19,284 @@
...
+    sec_trigger_delays: dictionary
+        the delays for the secondary channels that have caused a trigger.
+        If there is no trigger or no secondary channels, it's an empty dictionary
+    """
+    station_id = station.get_id()
+    beam_rolls = get_beam_rolls(station, det, triggered_channels, phasing_angles,
+                                ref_index=ref_index, trigger_adc=trigger_adc,
+                                upsampling_factor=upsampling_factor)
+
+    secondary_beam_rolls = [{} for direction in range(len(phasing_angles))]
```
clark2668, line 189: The syntax of [{} for direction in ...] is too pythonic imo, and makes it hard to understand what the line is trying to accomplish. Perhaps a comment would help.

NuRadioReco/modules/phasedarray/triggerSimulator.py
```diff
@@ -18,265 +19,284 @@
...
+                                              ref_index=ref_index, trigger_adc=trigger_adc,
+                                              upsampling_factor=upsampling_factor)
+
+    traces = {}
+    for channel_id in triggered_channels:
+        channel = station.get_channel(channel_id)
+        channel_id = channel.get_id()
+        time_step = 1 / channel.get_sampling_rate()
+
+        if trigger_adc:
```
clark2668, line 201: Would be nice to explain what is happening here. I think what what you're doing is getting all the traces pulled into a general array? But it would be good to just explain that so the reader isn't guessing.

NuRadioReco/modules/phasedarray/triggerSimulator.py
```diff
@@ -18,265 +19,284 @@
...
+                                          upsampling_factor=upsampling_factor,
+                                          nyquist_zone=nyquist_zone,
+                                          bandwidth_edge=bandwidth_edge)
+            time_step = 1 / det.get_channel(station_id, channel_id)['trigger_adc_sampling_frequency']
+            if upsampling_factor is not None:
+                upsampling_factor = int(upsampling_factor)
+                if upsampling_factor >= 2:
+                    time_step /= upsampling_factor
+            times = np.arange(len(trace), dtype=np.float) * time_step
+            times += channel.get_trace_start_time()
```
clark2668, line 218: a comment as to why this is necessary would be good

NuRadioReco/modules/phasedarray/triggerSimulator.py
```diff
@@ -18,265 +19,284 @@
...
-        return channel_trace_start_time
+        phased_trace = None
+        # Number of antennas: primary beam antennas + secondary beam antennas
+        Nant = len(beam_rolls[0]) + len(secondary_beam_rolls[0])
 
-    def check_vertical_string(self, station, det, triggered_channels):
-        """
-        Checks if the triggering antennas lie in a straight vertical line
-        """
+        for channel_id in traces:
```
clark2668, line 243: Explain more what is being done at a high level in the following lines.

NuRadioReco/modules/phasedarray/triggerSimulator.py
```diff
@@ -18,265 +19,284 @@
...
 
-            traces[channel_id] = trace[:]
+        # Implmentation of the ARA-like power trigger
+        window_width = int(window_time / time_step)
+        n_windows = int(len(trace) / window_width)
+        n_windows = 2 * n_windows - 1
+        squared_mean_threshold = Nant * threshold ** 2
 
-        for subbeam_rolls, sec_subbeam_rolls in zip(beam_rolls, sec_beam_rolls):
+        # Create a sliding window
```
clark2668, line 261: I think more explanation is needed to explain what the following lines are doing.

NuRadioReco/modules/phasedarray/triggerSimulator.py
```diff
@@ -18,265 +19,284 @@
...
 
-            for channel_id in traces:
-
-                trace = traces[channel_id]
-
-                if (phased_trace is None):
-                    phased_trace = np.roll(trace, subbeam_rolls[channel_id])
-                else:
-                    phased_trace += np.roll(trace, subbeam_rolls[channel_id])
+        if True in (squared_mean > squared_mean_threshold):
```
clark2668, line 269: Ditto about more explanation about what the following lines are doing.

NuRadioReco/modules/phasedarray/triggerSimulator.py
```diff
@@ -18,265 +19,284 @@
...
+                trigger_delays[channel_id] = subbeam_rolls[channel_id] * time_step
+            sec_trigger_delays = {}
+            for channel_id in sec_subbeam_rolls:
+                sec_trigger_delays[channel_id] = sec_subbeam_rolls[channel_id] * time_step
+            logger.debug("Station has triggered")
+            return True, trigger_delays, sec_trigger_delays
 
-                if (channel_id in sec_subbeam_rolls):
-                    phased_trace += np.roll(trace, sec_subbeam_rolls[channel_id])
+    return False, {}, {}
```
clark2668, line 279: Why is the default return value False?

NuRadioReco/modules/phasedarray/triggerSimulator.py
```diff
@@ -374,66 +394,51 @@ def run(self, evt, station, det,
...
-                                                                                          window_time, cut_times,
-                                                                                          trigger_adc=trigger_adc,
-                                                                                          upsampling_factor=upsampling_factor,
-                                                                                          nyquist_zone=nyquist_zone,
-                                                                                          bandwidth_edge=bandwidth_edge)
+            if ((not coupled) and (len(secondary_channels) > 0)):
+                primary_trigger, trigger_delays, dummy_delay = phased_trigger(station, det, threshold,
+                                                                              triggered_channels,
+                                                                              phasing_angles,
+                                                                              [],
```
clark2668, line 410: Imo for empty arguments, e.g. [], better to name the parameter for clarity, e.g. "secondary_channels=[]" That or re-work the phased_trigger() function so the arguments are truly optional and then just omit them?

clark2668 (changes requested):
Hi Christian--thanks for working on this! A few general comments:

- I think some of the naming is a bit opaque and could be improved. E.g., what is a "roll"? I think what you mean is "delays" so better to just say that. E.g. `get_beam_rolls` -> `get_beam_delays`

- A high-level discussion what what is actually being done in the code I think would help a lot, particularly for the for the "phased_trigger" function. Something like: "Make a copy of every trace for all the antenna. Then, delay each of those traces by the relevant delays for a given phasing angle. Then add the delayed traces together in voltage, and then square it before forming the trigger."

There's a few other things GitHub wouldn't let me comment on because they weren't formally part of the changeset:

L14-19: clarification on how the "main" and "secondary" angles are chosen should be added
L351: typo "if True not trigger" -> "if True no trigger"

Some default parameters could benefit from further explanation imo:
	L145: rather than "triggering channels" say "Id's of the channels to be included in the phasing"
	L310: how was 10.67ns chosen as the window_time? Should be explained.
	L312: ref_index should be the n(z) at what point? Center of the PA...? How will this be handled for a PA in the firn, where that is changing rapidly?
	L313: it should be further explain how "cut_times" should be chosen

cg-laser:
Hi Brian, thanks for the thorough check of the module. As I agree with you on most of the points, I won't be able to do something about it as I did not write this module. I just refactored it a bit, so that the noise trigger rate calculation can directly use this module (Before it was done by code duplication which I find quite error prone as any change needs to be implemented in two places)

clark2668:
Hi Christian--fair enough! I defer to and others on how to proceed then. The typos and function renaming is probably easy, but the "What is this actually trying to do" is probably harder.

cg-laser:
I think there is a problem in the script that calculates the trigger efficiency curves. The SNR is calculated on the unfiltered signal.

dansmithphysics:
I've been banging my head against the wall for a week and a half now making diagnostic plots. The issue is that the phased array isn't triggering at the single-dipole rate that it should be. The efficiency vs SNR plots (currently running, will upload later) don't look good when running over the same sim data sample that is used to create the Veff plots. The efficiency vs SNR plots look much better if I use boresight only data samples, which suggests somethings wrong with beaming. 

Screen Shot 2020-10-07 at 9 42 15 AM
Screen Shot 2020-10-07 at 9 40 59 AM
Screen Shot 2020-10-07 at 9 39 18 AM
Screen Shot 2020-10-07 at 9 38 46 AM
Screen Shot 2020-10-07 at 9 38 08 AM

dansmithphysics:
While the github linter continues to allude me, I think I've figured out what was wrong with my above problem. I wasn't correctly upsampling, so beams were extremely narrow. I thought things were working last week because I was making efficiency vs SNR plots using only boresight data, where beaming was 'correct'. As soon as I considered events evenly distributed in space, the holes between beams became clearer so dropped performance. With proper number of taps for upsampling, makes things much more reasonable. Attached plots are for a 100 m station. Besides fixing the linter code and commenting out some deprecated code, the phased array module is close to being done.

One thing of note, the RNO-G proposal has a figure made by Daniel that has a quoted sensitive curve quite a bit worse than what I have here. Only thing that comes to mind is that one of us got the noise level wrong. 

Screen Shot 2020-10-08 at 12 10 44 AM

Screen Shot 2020-10-08 at 12 09 40 AM

cg-laser:
What uptime did you assume to convert the Veff to expected sensitivity? And how many stations did you use? I expect this the main difference to the RNO-G white paper plot which assumed a realistic uptime (only solar power I think) and 5 years of operation

cg-laser:
I saw that you used an upsampling of 2 for the 4 channel setup, and an upsampling of 4 for the 8 channel setup. How much does the sensitivity change if a 4 channel setup is used without upsampling (which I thought was the baseline plan)? 
And just to be sure: The module is flexible enough to correctly calculate the beams for any upsampling and number of antennas?

dansmithphysics:
Good point about the livetime, I was assuming 100% livetime. I used 35 stations, 5 years. So the livetime would explain a factor of 2. I guess one question is with noise. If the digitizer goes up to say 1 GHz but we have a trigger that goes from 0-250 Mhz, which bandwidth should be used to calculate the noise, ~250 Mhz or ~1 GHz?

I believe the baseline plan for the 4 channel setup was going to be no upsampling with some secondary beaming, but the beams work out pretty well with an upsampling factor of 2. The module should be flexible enough to do any integer upsampling and number of antennas. I did strip away the code for the secondary beams, something I'm planning to add back in just in case. I also should fix any code (for example, envelope phase trigger) that depends on this triggering module.

cg-laser:
You need to use the bandwidth of the filter which should be more like ~140MHz (100 - 240MHz). You get the precise number by integrating over the filter.

dansmithphysics:
After a bit more debugging, I found I was misunderstanding how the noise gets added in the various functions that NuRadioMC call up (I redefined the _detector_simulation_part2 so filters in the _detector_simulation_filter_amp weren't getting called). There was also a small filtering bug that I removed. Attached are the most up to date results. 

The efficiency vs SNR still isn't behaving exactly as we expect for the single antenna equivalent. Note that, for the Veff plot, there is noise for the phased array but not the single antenna equivalent. I made a residual plot against the 1.5 sigma single dipole trigger to better see the behavior in the Veff plot. For certain energies, things agree with the efficiency vs SNR plot, but things perform worse at lower energies and better at high energies than we would expect from the single antenna. Not sure if this points to things not being properly coded or a breakdown of our understanding of the phased array. Only thing that comes to mind is planewavy-ness and spectral content vs energy. 

I also ran for two different integration window, and found a slight improvement using a shorter (16 ns) integration window. 

Screen Shot 2020-10-16 at 10 28 00 AM
Screen Shot 2020-10-16 at 10 24 24 AM
Screen Shot 2020-10-16 at 9 57 19 AM
Screen Shot 2020-10-16 at 9 52 42 AM
Screen Shot 2020-10-16 at 9 52 33 AM
Screen Shot 2020-10-16 at 9 52 21 AM

NuRadioReco/examples/PhasedArray/SNR_curves/T02RunSNR_RNOG.py
```diff
@@ -0,0 +1,280 @@
...
+    exit()
+
+def count_events():
+    count_events.events += 1
+count_events.events = 0
+
+class mySimulation(simulation.simulation):
+
+    def _detector_simulation_filter_amp(self, evt, station, det):
+        channelBandPassFilter.run(evt, station, det,
```
cg-laser, line 115: inconsistent definition of filter. Should be the same as in `detector_simulation_part2`

NuRadioReco/examples/PhasedArray/SNR_curves/T02RunSNR_RNOG.py
```diff
@@ -0,0 +1,280 @@
...
+        Vpps = np.zeros(self._station.get_number_of_channels())
+        for iChan, channel in enumerate(self._station.iter_channels()):
+            trace = np.array(channel.get_trace())
+            Vpps[iChan] = np.max(trace) - np.min(trace)
+
+        factor = 1.0 / (np.mean(Vpps) / 2.0)
+        mult_factors = factor * SNRs
+
+        has_triggered = True
+        while(has_triggered): # search for noise traces that don't set off a trigger
```
cg-laser, line 150: I don't get what you're doing here. You seem to loop until you found a noise fluctuation that trigger but then you don't do anything with it.

NuRadioReco/examples/PhasedArray/SNR_curves/T02RunSNR_RNOG.py
```diff
@@ -0,0 +1,280 @@
...
+                squared_mean_threshold = np.power(threshold, 2.0)
+
+                has_triggered = (True in (squared_mean > squared_mean_threshold))
+
+            if(has_triggered):
+                print('Trigger on noise... ')
+
+        filtered_noise_traces = {}        
+        for channel in self._station.iter_channels():
+            filtered_noise_traces[channel.get_id()] = channel.get_trace()
```
cg-laser, line 198: doesn't this set `filtered_noise_traces` to the noise fluctuation that triggered the phased array where the Askaryan signal is zero ?

NuRadioReco/examples/PhasedArray/SNR_curves/T02RunSNR_RNOG.py
```diff
@@ -0,0 +1,280 @@
...
+                                                     threshold = threshold,
+                                                     triggered_channels=channels,
+                                                     phasing_angles=phasing_angles,
+                                                     ref_index = 1.75,
+                                                     trigger_name='primary_phasing',
+                                                     trigger_adc=False,
+                                                     adc_output='voltage',
+                                                     trigger_filter=None,
+                                                     upsampling_factor=upsampling_factor,
+                                                     window=int(32 * new_sampling_rate * upsampling_factor / 2),
```
cg-laser, line 219: why do you divide by 2 here?

NuRadioReco/examples/PhasedArray/SNR_curves/T02RunSNR_RNOG.py
```diff
@@ -0,0 +1,280 @@
...
+
+            if(has_triggered):
+                print('Trigger for SNR', SNRs[iSNR])
+                SNRtriggered[iSNR] += 1
+
+        count_events()
+        print(count_events.events)
+        print(SNRtriggered)
+
+        if(count_events.events % 10 == 0):
```
cg-laser, line 241: why do you only save every 10th trigger? Seems inefficient to me.

cg-laser (commented):
great work, thanks a lot for the update. The plots are really useful to get a sense of the functionality. 

I went in detail through the code and have a few comments. Some of the are quite picky but I think it is important to have this module in a good shape as it will be the foundation for most future studies. I partly commented directly on the file itself (which should show up as a separate comment) as well as additional comments below (sorry for the mess). 

phased array trigger module:

description of `triggered_channels` missing in `get_beam_rolls` function

can we rename the function `get_beam_rolls` to a more descriptive name? Something like `calculate_time_delays`?

several functions have incomplete doc strings. All input parameters should be described in the doc strings. 

rename `powerSum` to `power_sum` to follow standard python naming scheme. 

Can you describe 
```
step: int
       Time step in power integral
```
better? If I understood it correctly, you move the integration window in discrete steps along the trace? And maybe add the info that the physical length of the window and the step will depend on the initial sampling rate of the detector as well as the upsampling factor. 

I'm wondering if it would make sense to change the definition of the trigger threshold. Currently `threshold` is specified by the user which is transformed to 
`squared_mean_threshold = Nant * threshold ** 2`
which is compared to the squared sum of the trace of `window` samples which is divided by `window`. Because we also determine the trigger threshold by simulating the thermal noise trigger rate with the same code, it is internally consistent. But maybe a more physical definition will be helpful? For example: The user defines the time integrated power threshold (with units of V^2 * time). The module could then calculate the integrated power by `sum(trace[window]**2) * Delta T` which will also have units of V^2 * time and can be directly compared to the value specified by the user. But both options are fine with me. 

Can you calculate the trigger thresholds with a higher statistical precision? You could require that you get at least 100 triggers for each threshold. 
Do I see it correctly that the thresholds needs to be recalculated for every change in detector sampling rate, upsampling, window, step, number of antennas and number of beams? The exact definition of the beam angles should be irrelevant as the noise is random. 

Do you use `T01MeasureNoiselevel.py` to calculate the thresholds that correspond to a certain noise level? I don't think the filters are quite right. We should use 
```
channelBandPassFilter.run(evt, station, det,
                                  passband=[1 * units.MHz, 230 * units.MHz], filter_type="cheby1", order=9, rp=0.1)
channelBandPassFilter.run(evt, station, det,
                                  passband=[80 * units.MHz, 800 * units.GHz], filter_type="cheby1", order=4, rp=0.1)
```

This calculation `window=int(16 * sampling_rate * upsampling_factor)` only works because the default time unit is `ns`. Better replace it by `window=int(16 * units.ns * sampling_rate * upsampling_factor)` if you want the window to be 16 nano seconds long. 

Can you remove the outdated `noise_trigger_rate.py` script?

NuRadioReco/examples/PhasedArray/SNR_curves/T02RunSNR_RNOG.py
dansmithphysics: It loops until it finds a noise trace that doesn't trigger the phased array, so that noise triggers don't slip into the efficiency plot. It is a bit confusing, it keeps going until has_triggered = false on a noise trace.

NuRadioReco/examples/PhasedArray/SNR_curves/T02RunSNR_RNOG.py
dansmithphysics: This sets filtered_noise_traces to the noise fluctuation that didn't trigger the phased array, and then the filtered askaryan signal gets added back in later on.

NuRadioReco/examples/PhasedArray/SNR_curves/T02RunSNR_RNOG.py
dansmithphysics: That is a hold-over from a previous integration window I was studying (32 ns vs 16 ns), I'll make sure to remove it for clarity.

NuRadioReco/examples/PhasedArray/SNR_curves/T02RunSNR_RNOG.py
dansmithphysics: That was for when I was debugging. What seems more efficient, waiting until the end?

PR #260, opened 2020-09-15
Lets channelGenericNoiseAdder store SNR

christophwelling:
I would like to be able to see the SNR calculated from the noiseless signal instead of signal + noise.
Since the channelGenericNoiseAdder knows both the noiseless trace and the added noise, it is the logical place to do this.


cg-laser:
this won't give you the correct answer as the filters are applied after noise is being added. 

Instead you can check the SimChannels for the max amplitude. (its actually a bit more complicated as you would need to sum over the different ray tracing solutions and showers to get the final trace)

christophwelling:
In my opinion, this limitation is fine. One just has to be aware of where this SNR comes from and what limitations it has. That is why I set the default to False, so the SNR is not used without knowing where it comes from.
Calculating it this way is close enough (you can just set the noise bandwidth accordingly) and doesn't have the overhead that calculating it from the simChannels would have.

cg-laser:
the results should be pretty off for the default settings. We run the simlation with an internal sampling rate of 5 GHz -> max_freq = 2.5GHz. We also add noise up to this frequency. However, the signal normally only extends to ~500MHz or so. Vrms is proportionally to sqrt(bandwidth) so you're a factor of sqrt(2.5)/sqrt(0.5) = 2 wrong. (Assuming you only filter up to 500MHz.

cg-laser:
If you really need this I suggest to implement it properly : Run the filter on a copy of the station before adding noise. Then run the `channelSignalReconstructor` on the station copy and save the relevant parameters (signal amplitude etc ) back to `self._station`. This can be make configurable via the config file.

christophwelling:
I think we are talking about 2 different use cases here. I don't intend to run the noise adder in the NuRadioMC simulation itself. I want to be able to change the noise without having to rerun all the simulations, so I first simulate without noise and then continue with the .nur files.
For running inside NuRadioMC, your suggestion makes sense, but this doesn't work for my case, since there is no predefined filter simulation. There is also not really a problem with the sampling rate, because the signal will be downsampled to the detector sampling rate. Of course, any filters running afterwards will change the SNR, but that is pretty obvious and true no matter how SNR is calculated.

christophwelling:
With the new simChannels, it is probably easier and more flexible to calculate SNR using those. I'm closing this PR.

PR #261, opened 2020-09-17
Hardcoded channel IDs

christophwelling:
In some places, channel IDs were hardcoded. This PR makes the code more flexible by reading the available channels from the detector description and picking one of them


cg-laser (approved):
very good, thanks for fixing this

PR #262, opened 2020-09-17
Adds function to get the amp response directly from the detector class

christophwelling:
It's annoying to always use the analog_components modules manually every time you need the amp response. This PR simplifies things

### DIFF
--- a/NuRadioReco/detector/ARIANNA/analog_components.py
+++ b/NuRadioReco/detector/ARIANNA/analog_components.py
@@ -160,3 +160,7 @@ def get_cable_response(frequencies, path=os.path.dirname(os.path.realpath(__file
     cable_phase = np.exp(1j * cable_phase)
 
     return cable_amp * cable_phase
+
+
+def get_available_amplifiers():
+    return ['100', '200', '300']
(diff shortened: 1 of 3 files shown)


cg-laser (changes requested):
yes this is super useful. 

For sake of modularity, shouldn't the `analog_components` return a list of amplifiers they implement? 
Also this is not save for possible name clashes. This would be solved by giving the function an additional parameter "experiment" or so which would be "ARIANNA" or "RNO_G". Or a list of experiments. The default could be `["RNO_G", "ARIANNA"]` for now.

christophwelling:
Having a list of amps in the analog components is a good idea, I will add that.
I'm not sure about having the experiment as a parameter. The detector class itself should have all the information it needs to find the right amplifier. Otherwise this is can easily cause issues if we want to use it inside modules, where we can't easily change how this is called from the outside.
Could somehow add the name of the experiment to the detector description?

christophwelling:
I changed it so we get the amp names from the analog_components file now. It now also raises an error if the amp name is not unique

cg-laser (approved):
looks good. 

There is one thing though that is not supported. For ARIANNA we already have individual amp measurements available that can be requested via `get_amplifier_response` if an individual. amp is specified. However, we need think through how to properly implement amp measurements for all detectors at some point. This will be too much for this PR

PR #263, opened 2020-09-21
Adding traces

christophwelling:
With the electric fields and simChannels split up into individual traces per shower, we need an easy way to get the sum of all these traces to see what it all looks like together.
With this PR, the + operator is defined for 2 BaseTrace objects to return a new BaseTrace object containing the sum of the two traces. I think using the + operator works well here, since it is convenient, doesn't require importing anything and returns what one would intuitively guess the sum of two trace objects should be.
For example, plotting the sum of two traces would look like this:
tr = trace_1 + trace_2
ax1.plot(tr.get_times(), tr.get_trace())
ax2.plot(tr.get_frequencies(), abs(tr.get_frequency_spectrum())

### DIFF
--- a/NuRadioReco/framework/base_trace.py
+++ b/NuRadioReco/framework/base_trace.py
@@ -1,7 +1,10 @@
 from __future__ import absolute_import, division, print_function
 import numpy as np
 import logging
-from NuRadioReco.utilities import fft
+import fractions
+from NuRadioReco.utilities import fft, trace_utilities
+import scipy.signal
+import copy
 try:
     import cPickle as pickle
 except ImportError:
@@ -120,7 +123,7 @@ class BaseTrace:
             length = (self._frequency_spectrum.shape[-1] - 1) * 2
         return length
 
-    def apply_time_shift(self, delta_t):
+    def apply_time_shift(self, delta_t, silent=False):
         """
         Uses the fourier shift theorem to apply a time shift to the trace
         Note that this is a cyclic shift, which means the trace will wrap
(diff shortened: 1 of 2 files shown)


cg-laser (changes requested):
this is an awesome feature. It will simplify many things. One could also think of adding helper functions to directly return the simulated trace of one channel of one event (then summing all different showers and ray tracing contributions). 

I have a few comments though. 
1) The code could use more inline documentation to be more understandable for someone without a lot of knowledge (e.g. say what the trace shape corresponds to
2) from 193 - 230 is a lot of code duplication. Is there a way to reduce that? 
3) Do I see it correctly that 193-230 ist just appending the two traces with zeros to have a trace length long enough to store the full combined trace? Couldn't you combine the outer if/else by just doing  `time_offset = np.abs(self.get_trace_start_time() - x.get_trace_start_time())`
And at the end you just shift the "late" trace backwards using the fourier shif theorem?

christophwelling:
I managed to remove the code duplication depending on which trace starts first. Calculating the trace length is a bit more subtle (though your way to calculate the start time difference is much easier, no idea how I missed that) since the traces don't necessarily have the same length. So it could be that the length of the first trace is larger than the length of the second trace + the difference in trace start time.
I added some more documentation, it should now be easier to follow what is going on.
Yes, having some way to return summed traces would be useful. For E-fields it is a bit subtle since direct and reflected E-fields will have different directions so you can't just add them, but having that for the hadronic and EM part of a CC electron neutrino interaction would be really useful.

NuRadioReco/framework/base_trace.py
```diff
@@ -148,3 +155,82 @@ def deserialize(self, data_pkl):
...
+        # samples that time difference corresponds to
+        time_offset = np.abs(x.get_trace_start_time() - self.get_trace_start_time())
+        i_start = int(round(time_offset * sampling_rate))
+        # We have to distinguish 2 cases: Trace is 1D (channel) or 2D(E-field)
+        # and treat them differently
+        if trace_1.ndim == 1:
+            # Calculate length the new trace needs to hold both input traces
+            trace_length = max(first_trace.shape[0], i_start + second_trace.shape[0])
+            # Make sure trace has an even number of samples
+            trace_length += trace_length % 2
```
cg-laser, line 213: how does adding half of the trace length make the number of samples even? Don't you want to check if it is odd and then add `1` to the trace length?

christophwelling: % is the modulo operator. It returns 0 if trace_length is even and 1 if it is odd. So that's basically already what I'm doing

cg-laser: ups yes of course, I confused it with `//`

cg-laser (approved):

PR #264, opened 2020-09-28
Trace utilities

christophwelling:
Adds some methods to make it easier to handle traces


cg-laser (commented):
changes look good to me. This simplifies a lot of things.

NuRadioReco/framework/base_trace.py
```diff
@@ -37,6 +37,35 @@ def get_trace(self):
             self._frequency_spectrum = None
         return self._time_trace
 
+    def get_filtered_trace(self, passband, filter_type='butter', order=10):
```
cg-laser, line 40: can you use the `get_filter` method of the channelBandpassFilter module instead? Then we only need to implement filters at one common place. The parameters `station_id`, `channel_id`, `det` can just be `None`

cg-laser (changes requested):
just one inline comment.

NuRadioReco/framework/base_trace.py
christophwelling: Done, while I was at it I also implemented this for the E-field filters. I decided to define the filters in an extra function outside the channelBandPassFilter class. Otherwise I would have had to create a channelBandPassFilter object every time I want to filter a trace and that would probably be pretty bad for performance

christophwelling: Oops, I accidentally create a loop in the imports. The filters are now in utilities. That should fix it

cg-laser (changes requested):
looks great, thanks Christoph. Just one thing. `filter` is a standard python function. So we should better not name the module just 'filter'. Can you rename it to `bandpass_filter` or something similar if you have a better idea.

christophwelling:
Good point. I renamed the file

cg-laser (approved):

PR #265, opened 2020-10-06
SimChannels for cosmic ray events

christophwelling:
Makes the efieldToVoltageConverterPerEfield consider signal travel times between channels for cosmic ray events, so simChannels can be used for CR as well

### DIFF
--- a/NuRadioReco/modules/efieldToVoltageConverter.py
+++ b/NuRadioReco/modules/efieldToVoltageConverter.py
@@ -112,6 +112,9 @@ class efieldToVoltageConverter():
                         index_of_refraction = ice.get_refractive_index(antenna_position[2], site)
                     else:  # signal is coming from above, so we take IOR of air
                         index_of_refraction = ice.get_refractive_index(1, site)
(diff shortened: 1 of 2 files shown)


cg-laser (changes requested):
Can you add comments into the code that explains why you need to do that for cosmic rays?

The logic only works if the antennas have the depth. It might be useful to implement better time delay calculation here (?). At least, can you add a comment that this is assumed?

christophwelling:
I don't understand what you mean. This is the exact same thing as in the efieldToVoltageConverter.

cg-laser:
Yes I know. The functionality is also limited in that module. For ARIANNA, all antennas are at the same depth, so it does not matter. But for other configurations it might matter. 
I'm not saying that we need to implement it in this PR, but it will be helpful to mention that limitation in a comment in the code. 
I would also mention, that for cosmic rays, we simulate just one efield for each station

christophwelling:
Ah, I didn't consider the different index of refraction between air and ice. I added a comment explaining this

cg-laser (approved):
thanks Christoph

Issue #266, opened 2020-10-07
Singleton requirement in detector makes it difficult to work with >1 detector files


clark2668:
**Describe the issue**
Because the detector class is declared as a singleton, it is impossible to load two detector files at once, which might be useful for standalone purposes. For example, you can't do something like this:

```
dets = []
list_of_files = ['file1.json', 'file2.json']
for file in list_of_files:
  det = detector.Detector(json_filename=file, antenna_by_depth=False)
  dets.append(det)
```

You end up getting two copies of the first file:

```
print(dets)
[<NuRadioReco.detector.detector.Detector object at 0x7fa818050c40>, <NuRadioReco.detector.detector.Detector object at 0x7fa818050c40>]
```
Which is not helpful, and actually kinda confusing as to why the behavior is like that

**Expected behavior**
You'd prefer to be allowed to instantiate an indefinite number of detector objects.

**Additional context**
N/A

cg-laser:
@christophwelling I just added you to this issue because I think to remember that you ran into a similar problem?

The general problem is that we want both options. In a simulation, we don't want to create a new detector instance all the time, therefore the singelton. But for other cases it might be benefitial to open several detectors at once.

christophwelling:
You can create a new detector object by calling the __new__ method explicitly:
`
detector = NuRadioReco.detector.detector.Detector.__new__(NuRadioReco.detector.detector.Detector)
detector.__init__(source='detector_file.json')
`
This is for example done in NuRadioRecoio, when we read a detector from a .nur file and create a new detector object for each event file.

clark2668:
Ahhhh, I see. Is that obvious? If so, then chalk it up to my ignorance and we can move on :) If not, can we put a note about this somewhere in the docs?

cg-laser:
I agree that is not obvious. I would not have know it myself. @christophwelling would it be possible to just add a helper function like `initialize_new_detector` to the detector.py file which then executes exactly the code you posted above?

christophwelling:
Yes, I agree it is not really obvious, I also had to google it at the time. It's one of those things that are built in python, but you basically never have a reason to use.
I created this PR: https://github.com/nu-radio/NuRadioReco/pull/267 to fix this. Now you can just pass a parameter to the creater to get a new instance

anelles:
This looks like the issue was fixed in #267 @christophwelling, right? If you confirm, can you close this issue?

christophwelling:
Yes, this can be closed.

PR #267, opened 2020-10-09
Optional singleton

christophwelling:
Addresses #266 
No you can pass create_new=True to a Singleton's __call__ method to explicitly create a new class instance

### DIFF
--- a/NuRadioReco/autodoc/pages/utilities.rst
+++ b/NuRadioReco/autodoc/pages/utilities.rst
@@ -31,4 +31,19 @@ domain. The functions are wrappers around the the numpy real Fourier transform
 frequencies).
 The spectrum is defined in volt/GHz instead of volt/bin, so that the values are
 independent of the sampling rate.
-  .. Important:: Always use these helper functions when doing Fourier transformations.
+
+.. Important:: Always use these helper functions when doing Fourier transformations.
+
+Metaclasses
+------------------------
+
+Singleton
+^^^^^^^^^^^^
+For some classes (for example, the Detector class), it is usually a good idea to only have
+one instance of it active at a time. This can be accomplished by assigning the Singleton
+metaclass to it. The metaclass supercedes the __call__ method
+(e.g. :code:`NuRadioReco.detector.detector.Detector()`) to check if an instance of the class
+already exists and return that instance instead of creating a new one.
+
+If you want to enforce the creation of a new class instance, you can overwrite this behavior
+by passing create_new=True as a parameter to the __call__ method.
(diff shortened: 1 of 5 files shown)


NuRadioReco/autodoc/pages/utilities.rst
```diff
@@ -31,4 +31,19 @@ domain. The functions are wrappers around the the numpy real Fourier transform
...
+Singleton
+^^^^^^^^^^^^
+For some classes (for example, the Detector class), it is usually a good idea to only have
+one instance of it active at a time. This can be accomplished by assigning the Singleton
+metaclass to it. The metaclass supercedes the __call__ method
+(e.g. :code:`NuRadioReco.detector.detector.Detector()`) to check if an instance of the class
+already exists and return that instance instead of creating a new one.
+
+If you want to enforce the creation of a new class instance, you can overwrite this behavior
+by passing create_new=True as a parameter to the __call__ method.
```
clark2668, line 49: is the create_new list as a param somewhere in the singleton class definition? Might be good if it were.

clark2668 (commented):
I made one comment about adding the create_new parameter to the class def, but otherwise I think this looks good, and can confirm it works in the case I was having trouble with.

NuRadioReco/autodoc/pages/utilities.rst
christophwelling: Yes, it's in the docstring of the metaclass as well

cg-laser:
I think it is not clear yet how to call a new detector instance (at least not to me). Can you provide an explicit example? And ideally also add it to the docstring of the detector class as this is the point where people will look first?

christophwelling:
> so e.g. `det = Detector(source='json', create_new=True)` will do the job?

Yes.  I also noticed that by default the __init__ is not listed in the rendered documentation. I made a commit to change that for all classes.

cg-laser:
I think this is good to go but can you add the new parameter to the docstring of the detector __init__ function. I think otherwise it will be very difficult to see that this parameter exists.

clark2668:
Just wanted to comment that I started using this, and it was very helpful. Also looks good to merge after you're happy with the docs. Thanks!

christophwelling:
Then can someone approve this, so we stick to the process?

anelles (approved):
Looks good.

cg-laser (changes requested):
you did not add the description of the new __init__ parameter directly to the init function of the detector (see my comment/request above)

christophwelling:
I added it. Looks like sphinx does not know how to handle metaclasses, that's a bit unfortunate.

cg-laser (approved):
thanks

PR #268, opened 2020-10-12
fix weighting if it is not set in highlevel files. Now it is performe…

lpyras:
Fix weighting if it is not set in highlevel files. The weighting is set due to the area one station covers. The area is calulated using a voronoi diagramm. The vertices of each voronoi cell is determined in the showerplane (for symmetry reasons). Then, the vertices are transfromed on ground and the area of each cell is calculated. The value corresponds to the weight set. The area of all cells toghether equal the area of the footprint.


NuRadioReco/modules/io/coreas/readCoREASStation.py
```diff
@@ -45,8 +46,23 @@ def run(self, detector):
...
+                    # import matplotlib.pyplot as plt
+
+                    # fig, ax = plt.subplots()
+                    # im = ax.scatter(positions[:,0], positions[:,1], c=weights)
+                    # fig.colorbar(im, ax=ax)
+                    # plt.gca().set_aspect('equal')
+                    # plt.show()
+
                 else:
                     weights = coreas.calculate_simulation_weights(list(corsika["highlevel"].values())[0]["antenna_position"])
```
cg-laser, line 67: this line will crash as the function expects a zenith and azimuth attribute

cg-laser (changes requested):
I'm not exactly sure what this change does and if it still calculates what was calculated before. Can't you just calculate the vxvxB positions and pass these positions to the `calculate_simulation_weights` function? Or did you change the functionality of the function?

NuRadioReco/modules/io/coreas/readCoREASStation.py
lpyras: ah sure, i will change it,

lpyras:
> I'm not exactly sure what this change does and if it still calculates what was calculated before. Can't you just calculate the vxvxB positions and pass these positions to the `calculate_simulation_weights` function? Or did you change the functionality of the function?

Before, the positions were calculated directly on ground. For more inclined showers, the voronoi function did not really work anymore and gave ugly result. (see file attached)
test_voronoi The orange dots are the vertices of each cell and the blue dots are the station. Now I calculated the vertices in the shower plane were the grid is smooth, then the vertices for each point are projected on ground. The weight is the area of each cell on ground. I still does the same but avoids the mess for inclined showers. If it is directly calulated in the shower plane the inclination is not taken into account.
voronoi

cg-laser (approved):
thanks for the additional explanation

PR #270, opened 2020-10-15
Measured noise adder

christophwelling:
New module to add noise from measurements stored in .nur files to a simulation. The module randomly selects a suitable noise event from the files and add the measured traces as noise.


NuRadioReco/modules/channelMeasuredNoiseAdder.py
```diff
@@ -0,0 +1,142 @@
...
+        if noise_station is None:
+            raise ValueError('Could not find suitable noise event in noise files after {} iterations'.format(self.__max_iterations))
+        for channel in station.iter_channels():
+            noise_channel = noise_station.get_channel(channel.get_id())
+            channel_trace = channel.get_trace()
+            if noise_channel.get_sampling_rate() != channel.get_sampling_rate():
+                noise_channel.resample(channel.get_sampling_rate())
+            noise_trace = noise_channel.get_trace()
+            if channel.get_number_of_samples() > noise_channel.get_number_of_samples():
+                self.logger.warning(
```
cg-laser, line 80: is a warning enough? Depending on how long the simulated trace is, this might lead to wrong results.

cg-laser (commented):
looks good to me. I'm wondering, is this module a preparation for future RNO-G data or do you use it with current ARIANNA data? 
A general problem is, that the simulated trace is often longer than the noise trace. Only after the trigger is calculated, the trace can be cut to the correct length. But this required adding the noise before the trigger calculation. I never figured out how to solve this problem.

christophwelling:
I so far I am using for ARIANNA. I wanted to check that the IFT electric field reconstruction works with actual noise (fortunately it does). I guess once RNO is up, we can also use it for that as well.
I am currently using the trigger on noiseless traces. We don't have a good way to distinguish between signal and noise events right now, so I don't want to have to deal with accidental noise triggers right now. I know this is not exactly realistic, but to test reconstruction methods it is good enough for now. So right now this module is supposed to be used after the channelLengthAdjuster.

NuRadioReco/modules/channelMeasuredNoiseAdder.py
christophwelling: I changed the get_noise_station method to only accept stations if the noise trace is at least as long as the data trace.

cg-laser (changes requested):
can you change the random generator to use the new default way in NuRadioMC? (see https://github.com/nu-radio/NuRadioMC/pull/250). It essentially boils down to
```
from numpy.random import Generator, Philox
rnd = Generator(Philox(seed))
```

I guess performance could be optimized by first searching for a while in the same file, but this would be difficult to implement. Probably not worth the effort. 

Can you add the info (in the general docstring) that the noise traces need to be at least as long as the signal trace and what the intended module sequence is (i.e. after the length adjuster module)

christophwelling:
Done

PR #272, opened 2020-10-30
Detector browser fix

christophwelling:
-Fixes a problem in the way the channel IDs were assigned in the detector drawing
-Allows you to set the assume_inf and channel_by_depth parameters depending on what the detector description needs


cg-laser (approved):
looks good to me, but I don't know the code good enough to check it in detail

PR #273, opened 2020-11-05
Astropy time save

christophwelling:
Storing the station_time as astropy objects can cause trouble if the astropy versions used to read and write the files are different. Therefore we just save the value and format of the station_time so we can recreate it when reading.


cg-laser (approved):
I think it looks good. Then internally astropy just uses a single double to save the time?

christophwelling:
Yes, but annoyingly this internal variable can have different meanings depending on the format. So this is the safer option

PR #274, opened 2020-11-16
Parameter error serialization

christophwelling:
Serializes the electric field parameter covariances so they are stored in .nur files.
Also improves the serialization of station parameter covariances so that files remain readable if parameters are removed.
The change to the station parameters is not backward compatible, but I don't think anyone was using them. Old files shoudl be readable anyway, just not the parameter errors.


cg-laser (approved):
thanks for fixing this

PR #276, opened 2020-12-10
Event browser

cg-laser:
fix association of channel ids with channel index. 

To mitigate potentially out of order channels in the detector description, the detector now returns a sorted list.

### DIFF
--- a/NuRadioReco/detector/detector.py
+++ b/NuRadioReco/detector/detector.py
@@ -546,7 +546,7 @@ class Detector(object):
         channel_ids = []
         for channel in self.__get_channels(station_id).values():
             channel_ids.append(channel['channel_id'])
-        return channel_ids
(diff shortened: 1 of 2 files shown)


cg-laser:
I ran test manually and they pass.

christophwelling (approved):
Looks good, though I'm wondering if it wouldn't be even better if station.iter_channels() returned the channels in order as well.

cg-laser:
I wonder if this is potentially dangerous. Because then the order is silently changed and would e.g. not correspond to the order we have in the hdf5 files. 
The change I made would in principle also effect if simulations were already produced and if the channel order is rerequsted by the detector. But this is not what one would do normally. The `iter_channels` method is used and then the `channel_id` is used to extract certain parameters.

PR #277, opened 2020-12-10
fix trigger time adjuster

cg-laser:
in case no trigger, the trigger time was set to 0 which screwed up the logic in the triggerTimeAdjuster module. I made the triggerTimeAdjuster module more robust and changed the trigger modules to leave the trigger time at `None` if no trigger was registered. 
I noticed that because the traces were not cut correctly.

### DIFF
--- a/NuRadioReco/test/test_examples.sh
+++ b/NuRadioReco/test/test_examples.sh
@@ -15,22 +15,22 @@ rm -rf *deg_12/
 rm output_file.hdf5
 rm output_snr.json
 # Envelope phased array
-cd ../../EnvelopePhasedArray/Effective_volume/
-python3 T01generate_event_list.py minimal
-python3 T02RunPhasedRNO_shortband.py --inputfilename minimal_eventlist.hdf5
-rm minimal_eventlist.hdf5
-rm output*hdf5
-cd ../Noise_trigger_rate
-python3 noise_trigger_rate.py --ntries 10
-# Alias phased array
-cd ../../AliasPhasedArray/SNR_study
-python3 T01_generate_events_simple.py --n_events 10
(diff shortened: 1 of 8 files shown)


cg-laser:
the tests on NuRadioReco run fine, but the tests on NuRadioMC fail because of differences in the maximum amplitudes because the traces are now cut to different parts. We will need to fix this in NuRadioMC.

NuRadioReco/test/test_examples.sh
```diff
@@ -15,22 +15,22 @@ rm -rf *deg_12/
...
-# Alias phased array
-cd ../../AliasPhasedArray/SNR_study
-python3 T01_generate_events_simple.py --n_events 10
-python T02SNRNyquist.py input_alias_SNR.hdf5 phased_array_100m_0.25GHz.json config.yaml out.hdf5 out.json --nyquist_zone 2 --upsampling_factor 4 --noise_rate 1
-rm input_alias_SNR.hdf5
-rm out.hdf5
-rm out.json
-cd ../Noise_trigger_rate
-python3 noise_trigger_rate.py --ntries 10
+#cd ../../EnvelopePhasedArray/Effective_volume/
```
christophwelling, line 18: I think you forgot to undo something here :-)

christophwelling (changes requested):
Fix the test script please, rest looks good. Thanks for fixing

NuRadioReco/test/test_examples.sh
cg-laser: no they don't work anymore due to the phased array refactoring. This needs to be fixed later in the refractor-pa branch

christophwelling (approved):

PR #287, opened 2021-01-27
Add quadslotv2

iplaisier:
- The RNO-G Hpol converter (xfdtd, csv files) is added to antennapattern.py 
- The latest Vpol and Hpol are  updated in the RNO-G detector description

### DIFF
--- a/NuRadioReco/detector/antennapattern.py
+++ b/NuRadioReco/detector/antennapattern.py
@@ -8,7 +8,7 @@ from scipy import constants
 import logging
 import pickle
 import csv
-import cmath 
+import cmath
 
 logger = logging.getLogger('NuRadioReco.antennapattern')
 
@@ -109,7 +109,7 @@ def parse_RNOG_XFDTD_file(path_gain, path_phases):
         path to gain file
     path_phases:
         path to phases file
-        
+
     Returns:
     ----------
     all paramters of the file
@@ -121,8 +121,6 @@ def parse_RNOG_XFDTD_file(path_gain, path_phases):
         thetas = []
         gain_theta = []
         gain_phi = []
-        f = None
-        tmp_phi0_lines = []
         csv_reader = csv.reader(fin, delimiter=',')
         line_count = 0
         for row in csv_reader:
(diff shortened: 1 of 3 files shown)


NuRadioReco/detector/antennapattern.py
```diff
@@ -97,6 +98,106 @@ def get_group_delay(vector_effective_length, df):
...
+        path to gain file
+    path_phases:
+        path to phases file
+        
+    Returns:
+    ----------
+    all paramters of the file
+    """""
+
+    with open(path_gain, 'r', encoding='mac_roman') as fin:
```
cg-laser, line 117: do you need to specify an enconding? It seems that this is mac specific.

cg-laser (changes requested):
I compared the new XFDTD import/conversion function with the existing ARA reader. It seems that the file format is quite similar, maybe the two parse functions could be combined? But probably not that important. 

Comparing the calculation of the effective length I find an additional "4" in the denominator. It seems correct (comparing with A.13 of the NuRadioReco paper) if the input file is the realized gain. 

Actually I think there are more things wrong. The "n" used in A.13 is the index of refraction for which the realized gain was calculated, i.e. n=1 as these simulations were done for air. But what should change is the wavelength. 
`wavelength = c / np.array(ff)` should be replaced with `wavelength = c / np.array(ff)/n`
and H should just be
`H_theta = wavelength  * (50 / (4 * np.pi * Z_0)) ** 0.5 * gain_theta ** 0.5 * np.exp(1j * phase_theta)`

The function makes a few specific assumptions for rescaling with n, it would be nicer to have this as an parameter for the conversion function. 

Can you add the assumption about the antenna orientation also to the doc string/

cg-laser:
link to NuRadioReco paper: https://doi.org/10.1140/epjc/s10052-019-6971-5

anelles:
@iplaisier, please address the comments. Otherwise, this will be left behind when merging.

anelles:
@cg-laser can you re-review this?

cg-laser (approved):
I pushed a commit that refactors the code a bit to make it clearer what the underlying assumptions are. @Ilse are these the changes that we discussed via slack and is the code that you used to make the latest revision of the Hpol antenna?

iplaisier:
There was still a bug in the Hpol converter of the latest model i send you in calculating the phases (what i mentioned 2 weeks ago). This updated in this version. Furthermore it's the same and all the changes we discussed are included.

PR #289, opened 2021-02-19
Channel time offset calculator

christophwelling:
Adds module to that calculates the expected signal time difference between channels (along with some other properties) for a given vertex position. It does this by calculating correlations between channels and a template to determine the most likely scenario.


anelles:
@lpyras and @shallmann I assigned you just to spread the load a bit on reviewing. Please try to take a look at this code before the end of next week.

shallmann:
@christophwelling looks fine to me. Three minor comments, that are not obvious to me right away:
1/ L102f: why is correlation_size = a + b + 1? if you calculate a xcorr you would indeed get a length of a + b - 1 no?
2/ station "101" is used hard coded. I should know that myself already, but do we assume it exists always by convention?
3/ L142: I guess you know what you are doing. It was not obvious right away to me, can you comment on what is calculated here? (antenna_response['theta'] + antenna_response['phi']) gives you a summed amplitude?

christophwelling:
1. Yes, it should be a + b -1. Fixed that.
2. No, hardcoding station 101 was not intentional. Thanks for spotting.
3. It just assumes that the electric field is equally polarized in theta and phi component. It's useful so we don't have to worry where the antenna is more sensitive. I added a comment.

NuRadioReco/modules/channelTimeOffsetCalculator.py
```diff
@@ -0,0 +1,159 @@
...
+            for sim_shower in event.get_sim_showers():
+                if sim_shower.has_parameter(shp.vertex):
+                    vertex_position = sim_shower.get_parameter(shp.vertex)
+                    break
+        else:
+            if station.has_parameter(stnp.nu_vertex):
+                vertex_position = station.get_parameter(stnp.nu_vertex)
+            elif station.has_parameter(stnp.vertex_2D_fit):
+                vertex_2d = station.get_parameter(stnp.vertex_2D_fit)
+                vertex_position = np.array([vertex_2d[0], 0, vertex_2d[1]])
```
lpyras, line 91: Why is there a 0 at the position one? probably right though.

lpyras (approved):
Looks good to me. Thanks for doing this.

NuRadioReco/modules/channelTimeOffsetCalculator.py
shallmann: Yes, I think putting it along the x axis is fine here

shallmann (approved):

NuRadioReco/framework/parameters.py
```diff
@@ -39,7 +39,13 @@ class channelParameters(Enum):
     cr_xcorrelations = 8  # dict of result of crosscorrelations with cr templates
     nu_xcorrelations = 9  # dict of result of crosscorrelations with nu templates
     signal_time = 10  # the time of the maximum amplitude of the envelope
-    noise_rms = 11  # the root mean square of the noise
+    signal_regions = 11     # list of start and end times of regions that likely contain a signal
+    noise_regions = 12      # list of start and end times of regions that likel do not contain any signals
+    noise_rms = 13  # the root mean square of the noise
```
cg-laser, line 44: isn't it dangerous to rename the index of the enums? Would't this screw up existing data files as existing noise_rms values would be interpreted as `signal_regions`? I think the rule is: Never change an index, only add or delete  indices.

cg-laser (changes requested):

cg-laser (approved):

NuRadioReco/framework/parameters.py
christophwelling: Yes, I shouldn't have done that.

PR #293, opened 2021-02-22
Adds option to specify max number of events per file in eventWriter

christophwelling:
Adds an option to the eventWriter that lets you specify how many events should be written into one file.
This is useful when running reconstructions, because events can be split evenly between jobs, so all take roughly the same time.


cg-laser (approved):
looks good, useful feature!

Issue #288, opened 2021-02-11
Issues with singleton class for detectors


clark2668:
**Describe the issue**
The singleton class `create_new` method doesn't work as intended. For example, in the following, the stations in the "deep detector" seems to inherit the locations of the shallow detector--even though I've passed `create_new=True`.

```
det_deep = detector.Detector(json_filename='pa_200m_2.00km.json', antenna_by_depth=False, create_new=True)
det_deep.update(Time.now())
deep_station_ids = det_deep.get_station_ids()

det_shallow = detector.Detector(json_filename='surface_4LPDA_PA_15m_RNOG_300K_1.00km.json', antenna_by_depth=False, create_new=True)

for d_id in deep_station_ids:
	loc_deep = det_deep.get_absolute_position(d_id)
	print("Station {}, Deep {}".format(d_id, loc_deep)) # should be the position of deep station, but it reports the shallow locations
```
The det files are attached: [det_files.zip](https://github.com/nu-radio/NuRadioReco/files/5966623/det_files.zip)

It _is_ the case that the class instances _are_ different: 

```
print(det_deep)
print(det_shallow)
print(det_deep)
<NuRadioReco.detector.detector.Detector object at 0x7fcb80f6b790>
<NuRadioReco.detector.detector.Detector object at 0x7fcb87a69ac0>
<NuRadioReco.detector.detector.Detector object at 0x7fcb80f6b790>
```
so there must be something subtle in the class internal working going wrong maybe?

**Expected behavior**
The singleton class should generate a new class instance, but doesn't seem to be doing so.

**Additional context**
Thought this had been fixed in #266 , but or maybe I'm just doing something wrong.

christophwelling:
I couldn't recreate the problem. I create the shallow and deep detector instances from your JSON files, and both seemed to return the correct positions. I tried it for a few stations IDs.
Can you post the full script and the console outputs? For example, trying to access det_shallow with this script should throw an error that the detector time is not set. You probably want to use the GenericDetector class anyway, so that all the stations have channels.

clark2668:
Hi, sorry for the delay here. Yes, good idea, let me attach the full files and script. What I'm attaching in the zip file is the script (go.py), and the results of running with and without L8 commented out. When L8 is left _in_, it prints the following:

```
Station 1, Deep [-10000. -12500.      0.]
Station 2, Deep [-10000. -11500.      0.]
Station 3, Deep [-10000. -10500.      0.]
[...]

PR #294, opened 2021-02-22
Prevent interference between Detector instances

christophwelling:
Turns the SerializationMiddleware from TinyDB from a global variable into an attribute of the Detector class. This makes sure that each detector instance gets its own SerializationMiddleware object, to prevent interference between Detector instances.

Fix for #288 

@cg-laser This breaks the script NuRadioReco/detector/sql_to_json.py
I don't understand what this script is for or if we even need it any more.

### DIFF
--- a/NuRadioReco/detector/detector.py
+++ b/NuRadioReco/detector/detector.py
@@ -33,10 +33,21 @@ class DateTimeSerializer(Serializer):
         return datetime.strptime(s, '%Y-%m-%dT%H:%M:%S')
 
 
-def buffer_db(in_memory, filename=None, serialization=None):
+def buffer_db(in_memory, filename=None):
     """
     buffers the complete SQL database into a TinyDB object (either in memory or into a local JSON file)
+    
+    Parameters
+    ----------
+    in_memory: bool
+        if True: the mysql database will be buffered as a tiny tb object that only exists in memory
+        if False: the mysql database will be buffered as a tiny tb object and saved in a local json file
+    filename: string
+        only relevant if `in_memory = True`: the filename of the json file of the tiny db object
+    
     """
+    serialization = SerializationMiddleware()
+    serialization.register_serializer(DateTimeSerializer(), 'TinyDate')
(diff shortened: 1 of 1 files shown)


clark2668:
Hi Christoph, thanks for taking care of this so quickly! I get the expected behavior now.

NuRadioReco/detector/detector.py
```diff
@@ -156,6 +152,8 @@ def __init__(self, source='json', json_filename='ARIANNA/arianna_detector_db.jso
             Can be used to force the creation of a new detector object. By default, the __init__ will anly create a new
             object of none already exists.
         """
+        self._serialization = SerializationMiddleware()
```
clark2668, line 155: Do you want a `if self._serialization = None` here? Otherwise I don't see the point in passing the serialization as an argument?

clark2668 (commented):
Had a question about passing serialization as an argument if you're not going to check for it.

anelles:
@clark2668 Can you review this? You already took a look and I think it is easiest, if you check this as reviewer. We are trying to get NuRadioReco cleaned up before the end of next week. Thanks!

clark2668:
Sure, I'll take a look! I think I only have one final point of confusion.

NuRadioReco/detector/detector.py
```diff
@@ -33,11 +33,7 @@ def decode(self, s):
         return datetime.strptime(s, '%Y-%m-%dT%H:%M:%S')
 
 
-serialization = SerializationMiddleware()
-serialization.register_serializer(DateTimeSerializer(), 'TinyDate')
-
-
-def buffer_db(in_memory, filename=None):
+def buffer_db(in_memory, filename=None, serialization=None):
```
clark2668, line 36: Can more docs be added on what these arguments do? In particular, the purpose of `in_memory` and `serialization`? I don't see where the `serialization` argument would be used except in the case where `in_memory=False`? SO just a bit more clarity there would be nice. Thanks!

PR #296, opened 2021-02-25
Fixing expected polarization calculation

leshanz:
Expected polarization calculation is changed from:

`exp_pol_angle = np.arctan2(exp_efield_onsky[2], exp_efield_onsky[1])`

to

`exp_pol_angle = np.arctan2(abs(exp_efield_onsky[2]), abs(exp_efield_onsky[1]))`

Namely, taking the absolute values of etheta and ephi before taking the arctan of them. Otherwise, for some cosmic ray events, the expected polarization would be negative, with some extreme ones being even less than -90deg. However, the polarization should stay between 0deg to 90deg.

### DIFF
--- a/NuRadioReco/modules/voltageToAnalyticEfieldConverter.py
+++ b/NuRadioReco/modules/voltageToAnalyticEfieldConverter.py
@@ -613,7 +613,7 @@ class voltageToAnalyticEfieldConverter:
         exp_efield = hp.get_lorentzforce_vector(zenith, azimuth, hp.get_magnetic_field_vector(site))
         cs = coordinatesystems.cstrafo(zenith, azimuth, site=site)
         exp_efield_onsky = cs.transform_from_ground_to_onsky(exp_efield)
-        exp_pol_angle = np.arctan2(abs(exp_efield_onsky[2]), abs(exp_efield_onsky[1]))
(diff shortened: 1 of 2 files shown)


christophwelling:
I don't get why a negative polarization angle does not make sense. In principle the electric field vector does have a sign, so negative angles can be well defined. Whether or not one can measure this sign is a different question, but in principle this is possible.
If negative angles are a problem, you can always take the absolute value of the angle or use the get_normalized_angle funtion from the radiotools. So I don't think this change is really useful.

cg-laser:
@christophwelling @iplaisier please correct me if I'm wrong, but doesn't the analyticEfieldConverter reconstruct just the absolute value of the amplitude of the phi and theta component? Then it makes sense to also do the same for the expectation. 
But in general I agree, the sign makes a difference. And I think you're right that just taking the absolute number of the angle yields the same result. So maybe better just leave it as it is.

iplaisier:
Yes that's right. The analyticvoltagetoefield converter is not sensitive to the sign of the polarization

cg-laser:
we decided to leave the prediction as it is, as the sign of the electric field contains information. Only the comparison with reconstructed data can use a different definition if e.g. the reconstruction is not sensitive to the sign of the efield.

